PR #72, opened 2022-10-14
DATA-534 Map Saving Localizing (SLAM)

jeremyhyde-viam:
Adding localization mode. Currently this is done in two areas, one by setting the localization flag and also by removing map updates and checks of number of keyframes.

JIRA Ticket: https://viam.atlassian.net/browse/DATA-534

Associated PR: https://github.com/viamrobotics/rdk/pull/1473 and also https://github.com/viamrobotics/rdk/pull/1548


slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -498,22 +484,10 @@ void SLAMServiceImpl::ProcessDataOffline(ORB_SLAM3::System *SLAM) {
...
-                    poseGrpc = tmpPose.inverse();
-                    if (nkeyframes != keyframes.size()) {
-                        currMapPoints = currMap->GetAllMapPoints();
-                    }
-                }
-            }
-            nkeyframes = keyframes.size();
+            UpdateMapAndPosition(SLAM, tmpPose);
+
+            // This log line is needed by rdk integration tests.
```
jeremyhyde-viam, line 489: added this log and print for consistency between offline/online, however its not currently being used in integration tests in offline mode

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -58,6 +60,8 @@ class SLAMServiceImpl final : public SLAMService::Service {
     double yamlTime;
     std::atomic<bool> offlineFlag{false};
     bool local_viewer_flag = false;
+    bool pure_localization_mode = false;
+    int n_key_frames = 0;
```
kkufieta, line 64: [q] Is `n_key_frames` something that is save to be and/or should be a global variable within the slam service? Is `n_key_frames` not used to count the number of keyframes in each submap depending on the map you're looking at? If I'm wrong, what does it stand for?

slam-libraries/viam-orb-slam3/orbslam_server_v1_test.cc
```diff
@@ -110,7 +110,7 @@ BOOST_AUTO_TEST_CASE(ParseAndValidateArguments_valid_config_no_map_rate_sec) {
                               "-data_rate_ms=200",  "-map_rate_sec="};
     SLAMServiceImpl slamService;
     utils::ParseAndValidateArguments(args, slamService);
-    BOOST_TEST(slamService.map_rate_sec.count() == chrono::seconds(0).count());
+    BOOST_TEST(slamService.map_rate_sec.count() == chrono::seconds(60).count());
```
kkufieta, line 113: [q] Why does this have to change? I would have assumed that it would have had to be equal to a value of `60` before, too?

kkufieta (commented):
A few nits and questions. Another question: did you make changes in the code of `ORB_SLAM3`?

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -524,6 +498,27 @@ void SLAMServiceImpl::ProcessDataOffline(ORB_SLAM3::System *SLAM) {
     return;
 }
 
+// Update map nd position when new image is successfully passed to the system
+void SLAMServiceImpl::UpdateMapAndPosition(ORB_SLAM3::System *SLAM,
+                                           Sophus::SE3f tmpPose) {
+    // Update pose every loop during tracking as well as map when n_key_frames
```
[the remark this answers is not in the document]
jeremyhyde-viam: was referring to the loop above in the online and offline functions but will remove due to confusion

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
JohnN193: Fine with removing the log, as we have the `Passing image to SLAM` log already. Integration tests for offline mode are already setup and only look to see if all images are completed currently

JohnN193 (approved):
LGTM, but seems like something weird happened to your ORB_SLAM3 repo, would make sure thats up to date before merging

slam-libraries/viam-orb-slam3/orbslam_server_v1_test.cc
jeremyhyde-viam: in this test no value is given and so it was defaulted to 0. this behavior is not desired now that localization is 0. We can either have it default to extremely high map rate value to avoid saving a map OR treat it like the data rate and throw an error.

jeremyhyde-viam: I'm in favor of option 2

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
jeremyhyde-viam: number of keyframes will change when you have a valid map update occur that introduces a new key frame. This will cause us to grab all the map points again so we have the most up-to-date map.

If delocalization occurs or a new map is transferred to, this will also change the number of keyframes and cause us to grab a the new data. However I did just realize that if the new map has the same number of key frames the grabbing the new map data would not occur. This was a problem with the past method as well so I have added an additional check for if the current map id has changed as well

slam-libraries/viam-orb-slam3/orbslam_server_v1_test.cc
kkufieta: hmn let me think this through: If no `map_rate_sec` is provided via RDK, that means something went wrong, correct? Since RDK always should be sending a value for `map_rate_sec`. If we run orbslam locally, though, and the user doesn't provide a value for `map_rate_sec`, I would assume that the behavior should be the same as if the user doesn't provide a value through the config - so setting it to a default value of `60` would make sense to me, since it is coherent with how we handle it through RDK, and since providing this parameter should be optional. Am I making sense?

kkufieta: Discussed this in person - agreed to throw an error

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -754,9 +747,14 @@ void ParseAndValidateArguments(const vector<string> &args,
 
     auto map_rate_sec = ArgParser(args, "-map_rate_sec=");
     if (map_rate_sec.empty()) {
-        map_rate_sec = "0";
+        throw runtime_error("No map data rate specified");
```
kkufieta, line 750: [nit] Mention `map_rate_sec` in the error message. Example: `"map_rate_sec value is required"`

slam-libraries/viam-orb-slam3/orbslam_server_v1_test.cc
```diff
@@ -104,13 +104,20 @@ BOOST_AUTO_TEST_CASE(
     BOOST_TEST(slamService.slam_mode == "rgbd");
 }
 
+BOOST_AUTO_TEST_CASE(ParseAndValidateArguments_valid_config_no_data_rate_msec) {
+    const vector<string> args{"-data_dir=/path/to", "-config_param={mode=rgbd}",
+                              "-port=20000",        "-sensors=color",
+                              "-data_rate_ms=",     "-map_rate_sec=60"};
+    const string message = "No camera data rate specified";
```
kkufieta, line 111: [nit] I think we should definitely mention the actual missing variable. In this case, I don't know what the "camera data rate" is referring to. Saying that `data_rate_ms` has to be defined gives clear instructions and takes out the guessing.

kkufieta (approved):
LGTM once nits are addressed

PR #73, opened 2022-10-17
[DATA-277] Update date/time format to match datamanager

JohnN193:
Update slam repo to use the time format that other services are using. Currently setup to still work with our previous time format. 

rdk PR: https://github.com/viamrobotics/rdk/pull/1504

https://viam.atlassian.net/browse/DATA-277


slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -778,13 +779,24 @@ double ReadTimeFromFilename(string filename) {
...
+        return myTime;
+    } else {
+        return (double)thisTime;
+    }
+}
 
-    double myTime = (double)thisTime + sub_sec;
-    return myTime;
+// Checks if using new or old time format
+void FindTimeFormat(string filename) {
```
JohnN193, line 795: function intended to check which format is being used. Can be removed after PR in rdk goes thru

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -90,6 +90,9 @@ void ParseAndValidateArguments(const vector<string> &args,
 // Converts UTC time string to a double value.
 double ReadTimeFromFilename(const string filename);
 
+// Checks if using new or old time format
+void FindTimeFormat(string filename);
```
JohnN193, line 94: removed after rdk passes

slam-libraries/viam-orb-slam3/orbslam_server_v1_main.cc
```diff
@@ -100,6 +101,8 @@ int main(int argc, char **argv) {
             return 1;
         }
     }
+    viam::utils::FindTimeFormat(
```
JohnN193, line 104: remove after rdk passes

slam-libraries/viam-cartographer/src/io/read_PCD_file.cc
```diff
@@ -38,6 +38,12 @@ ReadFile::timedPointCloudDataFromPCDBuilder(std::string file_path,
 
     // KAT NOTE: The file name format for the pcd files is assumed to be, e.g.:
     // rplidar_data_2022-02-05T01_00_20.9874.pcd
+
+    // LEAVING COMMENTED UNTIL READY TO TEST
+    // double time_delta = ReadTimeFromFilename(
```
JohnN193, line 43: Leaving this commented as I did not get the chance to test this fully with cartographer and am unsure if other spots  would need this

slam-libraries/viam-orb-slam3/orbslam_server_v1_main.cc
```diff
@@ -159,6 +162,10 @@ int main(int argc, char **argv) {
         slamService.StartSaveAtlasAsOsa(SLAM.get());
         slamService.ProcessDataOnline(SLAM.get());
         slamService.StopSaveAtlasAsOsa();
+        while (viam::b_continue_session) {
+            this_thread::sleep_for(chrono::microseconds(
+                viam::checkForShutdownIntervalMicroseconds));
+        }
```
kkufieta, line 168: [q] Why did you add this? I would think that in the case of `ProcessDataOnline` finishing, the user already tried to kill the program and `b_continue_session` would be `false` anyways?

slam-libraries/viam-cartographer/src/io/read_PCD_file.h
```diff
@@ -13,14 +13,19 @@
 #include "cartographer/sensor/timed_point_cloud_data.h"
 
 namespace viam {
+static const int filenamePrefixLength = 6;
 namespace io {
 
 class ReadFile {
    public:
+    std::string time_format = "%Y-%m-%dT%H:%M:%SZ";
```
kkufieta, line 21: [q] Why not have this inside the scope of `ReadTimeFromFilename` since that's where it's used?

slam-libraries/viam-cartographer/src/io/read_PCD_file.h
```diff
@@ -13,14 +13,19 @@
 #include "cartographer/sensor/timed_point_cloud_data.h"
 
 namespace viam {
+static const int filenamePrefixLength = 6;
```
kkufieta, line 16: Can you move this into the `viam::io::` namespace?

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -778,13 +779,24 @@ double ReadTimeFromFilename(string filename) {
...
     time_t thisTime = std::mktime(&dt);
+    auto sub_sec_index = filename.find(".");
+    if ((sub_sec_index != string::npos)) {
+        double sub_sec = (double)std::stof(filename.substr(sub_sec_index), &sz);
+        double myTime = (double)thisTime + sub_sec;
+        return myTime;
+    } else {
+        return (double)thisTime;
+    }
+}
```
kkufieta, line 792: Would it be possible to put this also into a `ReadTimeFromFilename` function, together with the `string time_format` definition above?

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -866,16 +878,18 @@ int FindFrameIndex(const std::vector<std::string> &filesRGB,
...
+                  std::gmtime(&t));
     // Save the current atlas map in *.osa style
-    string path_save_file_name =
-        path_to_dir + "/" + camera_name + "_data_" + timestamp + ".0000.osa";
-    return path_save_file_name;
+    auto find_underscores = time_format.find("_");
+    if (find_underscores != string::npos) {
+        return path_to_dir + "/" + camera_name + "_data_" + timestamp +
+               ".0000.osa";
+    }
```
kkufieta, line 891: [q] will this be removed later, too?

kkufieta (commented):
Nice work - I have a few questions

slam-libraries/viam-orb-slam3/orbslam_server_v1_main.cc
JohnN193: leftovers from some debugging that I forgot about, ty for catching

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
JohnN193: yeah if we want to remove all support for the previous timestamp then this will be removed as well

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
JohnN193: these are currently inside the ReadTimeFromFilename function

slam-libraries/viam-cartographer/src/io/read_PCD_file.h
JohnN193: because other methods will probably need to know the time format, such as saving the pbstream. Should move this out of `ReadfFile` and into `io`?

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
JohnN193: as for `string time_format` I do not want to move that into the function, since multiple functions depend on it

slam-libraries/viam-cartographer/src/io/read_PCD_file.cc
```diff
@@ -38,6 +38,12 @@ ReadFile::timedPointCloudDataFromPCDBuilder(std::string file_path,
...
+    //     myYAML.substr(initial_filename.find("_data_") +
```
tessavitabile, line 44: `myYAML` should be `initial_filename`.

slam-libraries/viam-cartographer/src/io/read_PCD_file.h
kkufieta: I think for now I'd add it to the function. If it needs to be pulled out in the future for other functions, it can happen then.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -866,16 +878,18 @@ int FindFrameIndex(const std::vector<std::string> &filesRGB,
...
+    return path_to_dir + "/" + camera_name + "_data_" + timestamp + "Z.osa";
```
tessavitabile, line 892: Is the "Z" not already included in the timestamp?

slam-libraries/viam-orb-slam3/orbslam_server_v1_test.cc
```diff
@@ -133,7 +133,7 @@ BOOST_AUTO_TEST_CASE(ReadTimeFromFilename) {
 }
 
 BOOST_AUTO_TEST_CASE(FindFrameIndex_Closest_no_files) {
-    const string configTimeString = "2022-01-01T01_00_00.0000";
+    const string configTimeString = "2022-01-01T01:00:00.0000";
```
tessavitabile, line 136: Why doesn't this use Z?

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -778,13 +779,24 @@ double ReadTimeFromFilename(string filename) {
 
     // Now we read from buffer using get_time manipulator
     // and formatting the input appropriately.
-    ss >> std::get_time(&dt, "%Y-%m-%dT%H_%M_%SZ");
-    double sub_sec =
-        (double)std::stof(filename.substr(filename.find(".")), &sz);
+    ss >> std::get_time(&dt, time_format.c_str());
     time_t thisTime = std::mktime(&dt);
+    auto sub_sec_index = filename.find(".");
+    if ((sub_sec_index != string::npos)) {
```
tessavitabile, line 785: Why will there not always be sub-seconds?

tessavitabile (commented):
Nice! A couple comments/questions.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
JohnN193: because the .osa files would save with .0000 before, the new time format in rdk would end up removing the sub second portion when loading from an apriori map. I can add a comment in the code to clarify this

slam-libraries/viam-orb-slam3/orbslam_server_v1_test.cc
JohnN193: good catch, forgot to include the Z when updating the tests

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
JohnN193: good catch, forgot to fix this here

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
tessavitabile: Does that mean this can go away? (Actually, do you think you could put TODOs over everything that will go away?)

JohnN193: no this one would need to stay, unless we wanted to remove sub seconds from the yaml/osa files entirely

tessavitabile: Ah, thanks

JohnN193:
Do we want to remove support for the old format/datasets after integration tests in rdk pass? Just so stuff like the hackweek5 dataset still works if we plan to keep using it

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
tessavitabile, line 785: [nit] remove extra parentheses

tessavitabile (approved):
LGTM mod nit. I don't think we need to continue to support the old format. If we want to, we can convert hackweek5 to the new format.

slam-libraries/viam-cartographer/src/io/read_PCD_file.cc
```diff
@@ -86,7 +86,9 @@ int ReadFile::removeFile(std::string file_path) {
 
 // Converts UTC time string to a double value.
 double ReadFile::ReadTimeFromFilename(std::string filename) {
-    std::string time_format = "%Y-%m-%dT%H:%M:%SZ";
+    // TODO: change time format "%Y-%m-%dT%H:%M:%SZ"
+    // https://viam.atlassian.net/browse/DATA-638
+    std::string time_format = "%Y-%m-%dT%H_%M_%S";
```
kkufieta, line 91: Change this back to the original:

```
std::string time_format = "%Y-%m-%dT%H:%M:%SZ";
```

kkufieta (approved):
LGTM mod changing back to original `time_format` in `read_PCD_file.cc`

PR #78, opened 2022-10-18
fix required status checks

tessavitabile:
It turns out that skipping jobs that are required blocks merge. I couldn't find a great workaround, so I'm skipping the individual steps.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ on:
 
 jobs:
   appimage:
-    uses: viamrobotics/slam/.github/workflows/appimage.yml@main
+    uses: viamrobotics/slam/.github/workflows/appimage.yml@fix_required_status_checks
     secrets:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -18,7 +18,7 @@ on:
 
 jobs:
   appimage:
-    uses: viamrobotics/slam/.github/workflows/appimage.yml@main
+    uses: viamrobotics/slam/.github/workflows/appimage.yml@fix_required_status_checks
```
tessavitabile, line 21: Will revert before merging

.github/workflows/pullrequest.yml
```diff
@@ -16,6 +16,6 @@ on:
 
 jobs:
   test:
-    uses: viamrobotics/slam/.github/workflows/test.yml@main
+    uses: viamrobotics/slam/.github/workflows/test.yml@fix_required_status_checks
```
tessavitabile, line 19: Will revert before merging

PR #85, opened 2022-10-24
[DATA-118] Add non-live (offline) ingestion loop

kkufieta:
https://viam.atlassian.net/browse/DATA-118

More work will be needed on the `ProcessDataOffline` function, but it was for the most part already finished under a different name.

### DIFF
--- a/slam-libraries/viam-cartographer/src/main.cc
+++ b/slam-libraries/viam-cartographer/src/main.cc
@@ -45,9 +45,6 @@ int main(int argc, char** argv) {
     std::unique_ptr<grpc::Server> server(builder.BuildAndStart());
     LOG(INFO) << "Server listening on " << *selected_port << "\n";
 
-    // Start SLAM: Placeholder function with full "offline mode" functionality
-    // that will be picked apart with future tickets into separate functions
-    // (GetMap, GetPosition, ProcessDataOnline, ProcessDataOffline).
     LOG(INFO) << "Start mapping: offline mode\n";
     slamService.ProcessDataOffline();
     LOG(INFO) << "Done mapping: offline mode\n";
--- a/slam-libraries/viam-cartographer/src/main.cc
+++ b/slam-libraries/viam-cartographer/src/main.cc
@@ -49,7 +49,7 @@ int main(int argc, char** argv) {
     // that will be picked apart with future tickets into separate functions
     // (GetMap, GetPosition, ProcessDataOnline, ProcessDataOffline).
     LOG(INFO) << "Start mapping: offline mode\n";
-    slamService.CreateMap();
+    slamService.ProcessDataOffline();
     LOG(INFO) << "Done mapping: offline mode\n";
 
     while (viam::b_continue_session) {
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -144,9 +144,8 @@ void SLAMServiceImpl::CreateMap() {
 
     if (this->starting_scan_number < 0 ||
         this->starting_scan_number >= int(file_list.size())) {
-        std::cerr << "starting_scan_number is out of bounds: "
-                  << this->starting_scan_number << std::endl;
-        return;
+        throw std::runtime_error("starting_scan_number is out of bounds: " +
+                                    std::to_string(this->starting_scan_number));
     }
 
     std::cout << "Beginning to add data....\n";
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -106,10 +106,26 @@ void SLAMServiceImpl::SetUpMapBuilder() {
     mapBuilder.BuildMapBuilder();
 }
 
-void SLAMServiceImpl::CreateMap() {
+void SLAMServiceImpl::ProcessDataOffline() {
     // Set up and build the MapBuilder
     SetUpMapBuilder();
 
+    // TODO: Check whether or not we're running localization and/or
+    // updating a map. Create a function to run those modes.
+    // Implemented with these tickets:
+    // https://viam.atlassian.net/browse/DATA-114
+    // https://viam.atlassian.net/browse/DATA-631
+
+    // TODO: Call localization only mode function, if localization only
+    // action mode
+
+    // TODO: Call updating map function, if apriori map was loaded
+
+    // Mapping a new map:
+    CreateMap();
+}
+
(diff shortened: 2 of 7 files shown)


slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -106,10 +106,26 @@ void SLAMServiceImpl::SetUpMapBuilder() {
     mapBuilder.BuildMapBuilder();
 }
 
-void SLAMServiceImpl::CreateMap() {
+void SLAMServiceImpl::ProcessDataOffline() {
```
kkufieta, line 109: `CreateMap` was for the most part already `ProcessDataOffline`. `ProcessDataOffline` will require more work with future tickets, but should satisfy the needs for this ticket for now.

slam-libraries/viam-cartographer/src/main.cc
```diff
@@ -49,7 +49,7 @@ int main(int argc, char** argv) {
     // that will be picked apart with future tickets into separate functions
```
tessavitabile, line 49: Should we remove this comment now?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -106,10 +106,26 @@ void SLAMServiceImpl::SetUpMapBuilder() {
...
+    // TODO: Call localization only mode function, if localization only
+    // action mode
+
+    // TODO: Call updating map function, if apriori map was loaded
+
+    // Mapping a new map:
+    CreateMap();
+}
+
+void SLAMServiceImpl::CreateMap() {
```
tessavitabile, line 128: Would it make sense to delete this function now, in favor of `ProcessDataOffline()`?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -128,9 +144,8 @@ void SLAMServiceImpl::CreateMap() {
```
tessavitabile, line 144: Is the result of `listFileInDirectory()` sorted?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -128,9 +144,8 @@ void SLAMServiceImpl::CreateMap() {
 
     if (this->starting_scan_number < 0 ||
```
tessavitabile, line 145: Are we using `starting_scan_number`, or should we remove this and always start from zero?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -128,9 +144,8 @@ void SLAMServiceImpl::CreateMap() {
...
         this->starting_scan_number >= int(file_list.size())) {
-        std::cerr << "starting_scan_number is out of bounds: "
-                  << this->starting_scan_number << std::endl;
-        return;
+        throw std::runtime_error("starting_scan_number is out of bounds: " +
+                                 std::to_string(this->starting_scan_number));
     }
 
     std::cout << "Beginning to add data....\n";
```
tessavitabile, line 151: Should we remove log.txt? Or keep it until we've implemented `GetPosition`?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -179,6 +194,8 @@ void SLAMServiceImpl::CreateMap() {
     mapBuilder.map_builder_->pose_graph()->RunFinalOptimization();
     PaintMap(mapBuilder.map_builder_, this->path_to_map + "/images", "0");
```
tessavitabile, line 195: Will we eventually get rid of the images folder (after we've implemented `GetMap`?

tessavitabile (commented):
Nice! A few comments and questions about how this will be changed in future tickets.

Also, did you test this through rdk?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I don't think so. `CreateMap` handles creating a map from scratch. When there is an apriori map, or when we're localizing only, the function looks differently. That's why I wrapped `CreateMap` within `ProcessDataOffline`, and we'll add if/else statements later on that'll execute a different function instead when updating/localizing.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: yes, it's sorted :)

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I think we should keep this for now, since we'll later connect this functionality once we implement `"input_file_pattern": "filepathglob"` (which is why we have it to begin with in the tech spec)

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I think there's no harm in keeping it until then. Might come in handy for debugging?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: Yes, we will!

kkufieta:
Yes, tested via RDK just now.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: sounds good!

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: sounds good!

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I am not sure. I think it's just not high enough priority right now, but I'd say the debate about it is still to be discussed. It's in the tech spec 🤷🏻‍♀️ And when we were testing, it was a helpful debugging tool. I can imagine it being useful to implement - especially for cartographer

tessavitabile: Sounds good! Added to the agenda for the SLAM meeting.

tessavitabile (approved):

PR #87, opened 2022-10-25
[DATA-159] Cartographer: Extract jpeg map image for GRPC endpoint

kkufieta:
https://viam.atlassian.net/browse/DATA-159

Implemented the jpeg grpc endpoint. Tested on:
* Raspberry Pi OS Lite (64-bit)
* 2.4 GHz 8-Core Intel Core i9; macOS Monterey


slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -46,64 +88,140 @@ SLAMServiceActionMode SLAMServiceImpl::GetActionMode() {
...
+            map_builder.OverwriteMaxSubmapsToKeep(max_submaps_to_keep);
+        }
+        if (slam_action_mode == SLAMServiceActionMode::UPDATING) {
+            map_builder.OverwriteFreshSubmapsCount(fresh_submaps_count);
+            map_builder.OverwriteMinCoveredArea(min_covered_area);
+            map_builder.OverwriteMinAddedSubmapsCount(min_added_submaps_count);
+        }
+        map_builder.OverwriteOccupiedSpaceWeight(occupied_space_weight);
+        map_builder.OverwriteTranslationWeight(translation_weight);
+        map_builder.OverwriteRotationWeight(rotation_weight);
```
kkufieta, line 108: Moved overwriting internal `MapBuilder` parameters the responsibility of `MapBuilder` itself, rather than letting the slam service do it. Found it also easier to read.

kkufieta (commented):
Ready for review.

slam-libraries/viam-cartographer/src/io/cairo_jpg.cc
```diff
@@ -0,0 +1,248 @@
+/* Copyright 2018 Bernhard R. Fischer, 4096R/8E24F29D <bf@abenteuerland.at>
```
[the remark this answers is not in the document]
tessavitabile: What's the reason to copy the source code, rather than including as a library? Is it because of the changes necessary to compile?

slam-libraries/viam-cartographer/src/io/image.h
```diff
@@ -0,0 +1,42 @@
+/*
```
tessavitabile, line 1: Remove cartographer copyright.

slam-libraries/viam-cartographer/src/io/image.h
```diff
@@ -0,0 +1,42 @@
...
+#include "cartographer/io/image.h"
+#include "cartographer/io/points_batch.h"
+
+namespace viam {
+namespace io {
+
+class Image : public cartographer::io::Image {
+   public:
+    using cartographer::io::Image::Image;
+    std::string WriteJpegMem(int quality);
```
tessavitabile, line 36: [nit] I would call this "WriteJpeg" or "WriteJpegToString".

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -104,6 +106,9 @@ class SLAMServiceImpl final : public SLAMService::Service {
     double translation_weight = 10.0;
     double rotation_weight = 1.0;
 
+    std::mutex map_builder_mutex;
+    mapping::MapBuilder map_builder;
```
tessavitabile, line 110: It's preferable to have a mutex and any variables it's protecting be private. This makes it easier to check that the mutex is locked before accessing the variables.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -72,6 +72,9 @@ class SLAMServiceImpl final : public SLAMService::Service {
     // MapBuilder parameters.
     void OverwriteMapBuilderParameters();
 
+    // PaintMap paints the map in jpeg format
+    void PaintMap(std::string &jpeg_img);
```
tessavitabile, line 76: [nit] It's more idiomatic to return a string than to use an out-parameter.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -46,64 +88,140 @@ SLAMServiceActionMode SLAMServiceImpl::GetActionMode() {
...
-        translation_weight);
-    mutable_ceres_scan_matcher_options->set_rotation_weight(rotation_weight);
 }
 
 void SLAMServiceImpl::SetUpMapBuilder() {
     auto action_mode = GetActionMode();
     if (action_mode == SLAMServiceActionMode::MAPPING) {
-        mapBuilder.SetUp(this->configuration_directory,
-                         this->configuration_mapping_basename);
+        {
```
tessavitabile, line 115: [nit] no need for extra curly braces

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -46,64 +88,140 @@ SLAMServiceActionMode SLAMServiceImpl::GetActionMode() {
...
+}
+
+void SLAMServiceImpl::PaintMap(std::string &jpeg_img) {
+    const double kPixelSize = 0.01;
+    cartographer::mapping::MapById<
+        cartographer::mapping::SubmapId,
+        cartographer::mapping::PoseGraphInterface::SubmapPose>
+        submap_poses;
+    {
+        std::lock_guard<std::mutex> lk(map_builder_mutex);
```
tessavitabile, line 149: If possible, I think we should lock map_builder only once and get everything we need from it, so that we get a consistent map.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -156,29 +278,27 @@ void SLAMServiceImpl::CreateMap() {
...
-                myfile << "global_pose: " << global_pose.DebugString()
-                       << std::endl;
-            }
-
-            if ((num_nodes >= this->starting_scan_number &&
-                 num_nodes < this->starting_scan_number + 3) ||
-                num_nodes % this->picture_print_interval == 0) {
-                PaintMap(mapBuilder.map_builder_, this->path_to_map + "/images",
-                         std::to_string(num_nodes));
+                auto local_poses = map_builder.GetLocalSlamResultPoses();
```
tessavitabile, line 293: What's the difference between GetLocalPoses() and GetLocalSlamResultPoses()?

tessavitabile (commented):
Nice! I meant to ask on your last PR: when you tested through rdk, were you just testing running cartographer through rdk (but using an existing dataset), or did you also test on images produced by rdk?

slam-libraries/viam-cartographer/src/io/cairo_jpg.cc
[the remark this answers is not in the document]
kkufieta: Tbh, I don't know how I would include it as a library. How could I do that?

The only other option I thought of was to import the repo as a git submodule but I didn't feel like it is worth the extra work of doing that - as compared to copying/pasting the two files it contains. This is the repo: https://github.com/rahra/cairo_jpg

Furthermore, adding those two files into `src/io` makes sense organization-wise. Also, I had to delete all other functions from the `cairo_jpg.cc` files in order for it to compile. I figured that leaving the copyright notice with the name of the person who wrote the code would be sufficient to give credit to them.

Happy to discuss more and hear what the options and best way to deal with this are.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I understand the concern and share it. The problem with `PaintMap` is that it takes a very long time to create the map: Between 2 - 30 seconds (see this ticket:https://viam.atlassian.net/browse/DATA-661). Locking `map_builder` for the entire time would mean mapping would be stopped for this long, which we can't do.

I investigated whether or not the map might change **while** we're painting it, and from what I understand and what I've seen in the practical tests, I believe that the map won't change during the printing process (dramatically). 

Given that currently map painting takes too long in order for us to block `map_builder` the entire time, and that we have a ticket for investigating if we can reduce this time, I have created another ticket that'll investigate whether or not we should block `map_builder` during the entire time while painting the map - or if we can find a way to extract all necessary information at once and THEN paint the map: https://viam.atlassian.net/browse/DATA-668

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: `GetLocalPoses` was simply a function within `server_functions.cc` that called `GetLocalSlamResultPoses()` and returned those results. Since we have `map_builder` now within the slam service, we can just call the function directly. I removed `GetLocalPoses` by deleting `server_functions.cc`.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
kkufieta: Initially, I returned a string but then thought that using the out-parameter would improve performance (which is really bad in the case of`PaintMap`) by removing one more copying step of the string. Am I wrong about this assumption? How can I implement `PaintMap` here by returning a string in a way that optimizes performance?

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
kkufieta: Why/how does it make it easier to check that the mutex is locked before accessing the variables when it is private?

Initially, I had it private but then ran into problems when I was using it. I'll take another look to see if I can make it private after all.

kkufieta (commented):
Thanks for the helpful review! :) To your question:

> I meant to ask on your last PR: when you tested through rdk, were you just testing running cartographer through rdk (but using an existing dataset), or did you also test on images produced by rdk?

I tested only on an existing dataset, because using images produced by rdk (i.e. running it in live mode, right?) is not implemented yet. The ticket to make this an option is: https://viam.atlassian.net/browse/DATA-609

Did I understand and answer your question correctly?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: Ah, right, this became an issue when I am calling `slamService.map_builder...` in the `slam_service_test.cc`, like this:

```cpp
void checkCartoMapBuilderParameters(SLAMServiceImpl& slamService) {
    auto pose_graph_options =
        slamService.map_builder.map_builder_options_.pose_graph_options();
...
```

resulting in this error:

```bash
/Users/kat/slam/slam-libraries/viam-cartographer/src/slam_service/slam_service_test.cc:13:21: error: 'map_builder' is a private member of 'viam::SLAMServiceImpl'
        slamService.map_builder.map_builder_options_.pose_graph_options();
```

Do you have suggestions on how I can resolve this elegantly? 

Something that comes to my mind is that I could implement functions within the `slamService` that retrieve & return those parameters from `map_builder`. Would this be the way to go here? Or are there alternative solutions?

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
tessavitabile: If the mutex and the variables it protects are private, then we only need to read the implementation of the class in order to check that the mutex is locked before accessing the variables. Otherwise, we have to look for usages of the mutex and variables outside of the class, so there's just a lot more to verify. Keeping the mutex private also makes it easier to reason about mutex acquisition order.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
tessavitabile: If `PaintMap` constructs a string and then returns it at the end of the function, the string will not be copied when it is returned, due to return value optimization (https://en.cppreference.com/w/cpp/language/copy_elision).

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: That is a long time! I agree we'll need to investigate it.

When I wrote my comment, I was imagining having this ticket do the work of extracting the necessary information from `map_builder` under the lock and then paint the map. It looks like that's what you described in DATA-668. The reason I thought we should do it now is because the behavior may not be well-defined when accessing an inconsistent view of the map. For example, the set of submap poses could change between when we get their IDs on line 142 and when we call `SubmapToProto()` on line 155 (though you do catch any errors in case an ID no longer exists). But if you've found that in practice, the map has changed little, and we're usually able to get a map without error that looks correct, then maybe it's fine to defer this investigation until DATA-668.

In the long-term (perhaps this is for DATA-661), I wonder if we should use cartographer's map-painting capability at all. `GetMap()` will need to be able to return a pointcloud as well, so we could transform the pointcloud into a jpeg in the same way that orbslam does. We would lose out on cartographer's pretty maps, but maybe that would be faster. And in the even longer-term, we won't be making jpegs within the slam libraries at all, since they will always return octrees, and the octrees will get converted to jpegs in rdk, so maybe we shouldn't spend too much time worrying about this.

tessavitabile: In answer to your second comment, yes, you could create getters for each of the `map_builder` parameters. A couple other options are:
- Add a function like `CheckMapBuilderParameters()` that takes in the expected parameters and compares them to the actual `map_builder` parameters
- Add a function `GetMapBuilderForTesting()` that returns the underlying `map_builder`. This should only be used in tests and the test is responsible for any concurrency requirements

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
kkufieta: Ah, right, regarding why I moved it to be public:

Having it private caused issues when I am calling `slamService.map_builder...` in the `slam_service_test.cc`, like this:

```cpp
void checkCartoMapBuilderParameters(SLAMServiceImpl& slamService) {
    auto pose_graph_options =
        slamService.map_builder.map_builder_options_.pose_graph_options();
...
```

resulting in this error:

```bash
/Users/kat/slam/slam-libraries/viam-cartographer/src/slam_service/slam_service_test.cc:13:21: error: 'map_builder' is a private member of 'viam::SLAMServiceImpl'
        slamService.map_builder.map_builder_options_.pose_graph_options();
```

Do you have suggestions on how I can resolve this elegantly? 

Something that comes to my mind is that I could implement functions within the `slamService` that retrieve & return those parameters from `map_builder`. Would this be the way to go here? Or are there alternative solutions?

tessavitabile:
> I tested only on an existing dataset, because using images produced by rdk (i.e. running it in live mode, right?) is not implemented yet. The ticket to make this an option is: https://viam.atlassian.net/browse/DATA-609. Did I understand and answer your question correctly?

Yes, that was my question. I think it would be better to test on images produced by rdk sooner rather than later. I think we could do it in offline mode by collecting a dataset from rdk (e.g. with `StartSlamProcess()` commented out). We might encounter issues with images produced by rdk, and it would be good to find out, especially if we need help from the Vision team. That sounds okay to wait until DATA-609, but good that we are going to do DATA-609 soon.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I agree with the idea of not using cartographer's map-painting capability - especially due to the color issues it causes. I believe we could implement our own version (similar to what you suggest with how orbslam implements it, though I would still do it similarly to how PaintMap does it, rather than through the point cloud, since important information would be lost that way). I do think we could speed map painting up that way and solve the color issues at the same time. It's good you mention the longer-term plan of using octrees to create a jpeg - which does really mean it's not worth putting too much effort into this.

Let me escalate the ticket that'll solve the issue we're discussing here to one of the top ones, so this will be fixed in a timely manner.

slam-libraries/viam-cartographer/src/io/read_PCD_file.h
```diff
@@ -15,16 +15,16 @@
...
+
+std::vector<std::string> ListFilesInDirectory(std::string data_directory);
+
+cartographer::sensor::TimedPointCloudData TimedPointCloudDataFromPCDBuilder(
+    std::string file_path, double start_time);
+
+int RemoveFile(std::string);
+
+// Converts UTC time string to a double value.
+double ReadTimeFromFilename(std::string filename);
```
kkufieta, line 28: Decided to take functions out of the class because they don't really share any variables and are just a collection of functions. Easier to understand where they're coming from since we now have to simply use the namespace `viam::io::FUNCTION`

kkufieta (commented):
Fixed PR comments + fixed a bug + cleaned up a little

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -154,49 +275,96 @@ void SLAMServiceImpl::CreateMap() {
...
+        map_builder.map_builder_->pose_graph()->RunFinalOptimization();
+    }
 
     LOG(INFO) << "Finished processing offline images";
 
     return;
 }
 
+int SLAMServiceImpl::GetOptimizeEveryNNodesFromMapBuilder() {
+    return map_builder.GetOptimizeEveryNNodes();
```
tessavitabile, line 323: These functions need to lock `map_builder`.

tessavitabile (approved):
LGTM mod comment

PR #96, opened 2022-10-28
[DATA-609] Add live ingestion loop

tessavitabile:
https://viam.atlassian.net/browse/DATA-609

This took less time than I thought, thanks to the work that had already been done! I tested both online and offline processing using data generated through rdk. I tested on a raspberry pi with the rplidar server running. I used this config for online (modified for offline):
```
{
  "services": [{
		  "name": "testcarto",
		  "type": "slam",
		  "attributes": 
        {
          "algorithm": "cartographer",
          "sensors": ["rplidar"],
          "config_params": {
          "mode": "2d",
          "v": "1"
        },
      "data_dir": "$HOME/carto-data"
    }}],
  "network": {
    "bind_address": "0.0.0.0:8080"
  },
  "remotes": [{
    "address": "localhost:4444",
    "name": "lidar",
    "prefix": false
  }]
}
```
Cartographer was able to process the data and respond to `GetMap` requests.

### DIFF
--- a/slam-libraries/viam-cartographer/src/io/read_PCD_file.cc
+++ b/slam-libraries/viam-cartographer/src/io/read_PCD_file.cc
@@ -85,9 +85,13 @@ int RemoveFile(std::string file_path) {
 
 // Converts UTC time string to a double value.
 double ReadTimeFromFilename(std::string filename) {
-    // TODO: change time format "%Y-%m-%dT%H:%M:%SZ"
-    // https://viam.atlassian.net/browse/DATA-638
-    std::string time_format = "%Y-%m-%dT%H_%M_%S";
+    std::string time_format = "%Y-%m-%dT%H:%M:%SZ";
+
+    // TODO DATA-638 remove this check
+    if (filename.find("_") != std::string::npos) {
+        time_format = "%Y-%m-%dT%H_%M_%S";
+    }
+
     std::string::size_type sz;
     // Create a stream which we will use to parse the string
     std::istringstream ss(filename);
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -218,7 +218,7 @@ std::string SLAMServiceImpl::PaintMap() {
     return jpeg_img;
 }
 
-void SLAMServiceImpl::ProcessDataOffline() {
+void SLAMServiceImpl::ProcessData() {
     // Set up and build the MapBuilder
     SetUpMapBuilder();
 
@@ -237,6 +237,43 @@ void SLAMServiceImpl::ProcessDataOffline() {
     CreateMap();
 }
 
+std::string SLAMServiceImpl::GetNextDataFile() {
+    if (offlineFlag) {
+        if (!b_continue_session) {
+            return "";
+        }
+        if (file_list_offline.size() == 0) {
+            file_list_offline = viam::io::ListFilesInDirectory(path_to_data);
+        }
+        if (file_list_offline.size() == 0) {
+            throw std::runtime_error("no data in data directory");
+        }
+        if (current_file_offline == file_list_offline.size()) {
+            LOG(INFO) << "Finished processing offline images";
+            return "";
+        }
+        const auto to_return = file_list_offline[current_file_offline];
+        current_file_offline++;
+        return to_return;
+    }
+    // online processing
+    while(b_continue_session) {
+        const auto file_list_online = viam::io::ListFilesInDirectory(path_to_data);
+        if (file_list_online.size() > 1) {
+            // Get the second-most-recent file, since the most-recent file may
+            // still be being written.
+            const auto to_return = file_list_online[file_list_online.size()-2];
+            if (to_return.compare(current_file_online) != 0) {
+                current_file_online = to_return;
+                return to_return;
+            }
+        }
+        VLOG(1) << "No new files found";
+        std::this_thread::sleep_for(data_rate_ms);
+    }
+    return "";
+}
+
 void SLAMServiceImpl::CreateMap() {
     cartographer::mapping::TrajectoryBuilderInterface *trajectory_builder;
     int trajectory_id;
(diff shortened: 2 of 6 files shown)


slam-libraries/viam-cartographer/src/io/read_PCD_file.cc
```diff
@@ -85,9 +85,13 @@ int RemoveFile(std::string file_path) {
 
 // Converts UTC time string to a double value.
 double ReadTimeFromFilename(std::string filename) {
-    // TODO: change time format "%Y-%m-%dT%H:%M:%SZ"
-    // https://viam.atlassian.net/browse/DATA-638
-    std::string time_format = "%Y-%m-%dT%H_%M_%S";
+    std::string time_format = "%Y-%m-%dT%H:%M:%SZ";
```
tessavitabile, line 88: I had to make this the default, since rdk is producing this format.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -253,34 +292,24 @@ void SLAMServiceImpl::CreateMap() {
...
-
-    std::cout << "Beginning to add data....\n";
+    LOG(INFO) << "Beginning to add data...";
 
     std::ofstream myfile;
     myfile.open("log.txt");
 
-    int end_scan_number = int(file_list.size());
-    for (int i = starting_scan_number; i < end_scan_number; i++) {
-        if (!b_continue_session) return;
```
tessavitabile, line 279: This is a slight behavior change, where if shutdown is signaled, we will still write the pbstream before exiting. However, I expect we will be removing writing the pbstream at shutdown, once we implement `map_rate_sec`.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -237,6 +237,45 @@ void SLAMServiceImpl::ProcessDataOffline() {
...
+            return "";
+        }
+        if (file_list_offline.size() == 0) {
+            file_list_offline = viam::io::ListFilesInDirectory(path_to_data);
+        }
+        if (file_list_offline.size() == 0) {
+            throw std::runtime_error("no data in data directory");
+        }
+        if (current_file_offline == file_list_offline.size()) {
+            LOG(INFO) << "Finished processing offline images";
```
tessavitabile, line 252: This is a slight behavior change, where we log this line before writing the pbstream. However, I expect we will be removing writing the pbstream at shutdown, once we implement `map_rate_sec`.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -121,12 +125,14 @@ class SLAMServiceImpl final : public SLAMService::Service {
     double rotation_weight = 1.0;
 
    private:
-    int starting_scan_number = 0;
```
tessavitabile, line 124: I decided to remove this for now, but we can add it back if/when we implement `input_file_pattern`.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -253,34 +292,24 @@ void SLAMServiceImpl::CreateMap() {
...
-        starting_scan_number >= int(file_list.size())) {
-        throw std::runtime_error("starting_scan_number is out of bounds: " +
-                                 std::to_string(starting_scan_number));
-    }
-
-    std::cout << "Beginning to add data....\n";
+    LOG(INFO) << "Beginning to add data...";
 
     std::ofstream myfile;
     myfile.open("log.txt");
```
JohnN193, line 298: only question is do we still want this? Not sure what it is/where its going

JohnN193 (approved):
LGTM, just a semi random question

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: @kkufieta was interested in keeping this for now.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -237,6 +237,45 @@ void SLAMServiceImpl::ProcessDataOffline() {
     CreateMap();
 }
 
+std::string SLAMServiceImpl::GetNextDataFile() {
+    if (offlineFlag) {
```
kkufieta, line 241: [nit] Can you put the code in this code block within `if (offlineFlag)` into a function? Would make it easier for my brain to process it with two sub-functions.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -237,6 +237,45 @@ void SLAMServiceImpl::ProcessDataOffline() {
...
+        }
+        if (current_file_offline == file_list_offline.size()) {
+            LOG(INFO) << "Finished processing offline images";
+            return "";
+        }
+        const auto to_return = file_list_offline[current_file_offline];
+        current_file_offline++;
+        return to_return;
+    }
+    // online processing
```
kkufieta, line 259: [nit] Similarly to above, can you put this into a function?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: It's just helpful example code for the "Extract pose for grpc endpoint" ticket: https://viam.atlassian.net/browse/DATA-158 

With which it should be removed for sure.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -314,8 +347,6 @@ void SLAMServiceImpl::CreateMap() {
         map_builder.map_builder_->pose_graph()->RunFinalOptimization();
     }
 
-    LOG(INFO) << "Finished processing offline images";
```
kkufieta, line 317: Keep the log but change the text to something like "Finished optimizing final map"

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -121,12 +125,14 @@ class SLAMServiceImpl final : public SLAMService::Service {
...
     int picture_print_interval = 50;
```
kkufieta, line 128: If you want to remove anything related to what will be implemented with `input_file_pattern`, then `picture_print_interval` should probably be deleted, too.

kkufieta (changes requested):
Nice work! Only a few comments and nits

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: Yes, we did remove writing a final map at shutdown from orbslam. However, in offline mode, orbslam will write a map immediately once it's processed all data: https://github.com/viamrobotics/slam/blob/392d8bcdc1b122c35d0a5f8c15b2b2a2eb2fd0f3/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc#L600-L602.

Oh, interesting about final map saving for cartographer. That sounds fine to do optimization at shutdown for cartographer. I just suggested we remove the final map saving from orbslam, since it wasn't in the spec and there was a bug we either needed to fix or just remove the code. I wonder how we'll handle final optimization for cartographer when we no longer save maps locally as files, but instead return them for cloud storage as a response to `GetInternalState`. Maybe we'll need to run final optimization in the cloud.

Sounds good, I'll add the log line you suggested.

PR #99, opened 2022-10-31
[DATA-158] pose grpc endpoint

JohnN193:
added grpc endpoint for pose based off the `log.txt` file that was setup previously and the orbslam grpc server(ty jeremy). https://viam.atlassian.net/browse/DATA-158

### DIFF
--- a/slam-libraries/viam-cartographer/src/mapping/map_builder.cc
+++ b/slam-libraries/viam-cartographer/src/mapping/map_builder.cc
@@ -96,25 +96,11 @@ cartographer::sensor::TimedPointCloudData MapBuilder::GetDataFromFile(
 
 // TODO: There might still be a lot of room to improve accuracy & speed.
 // Might be worth investigating in the future.
-cartographer::transform::Rigid3d MapBuilder::GetLastGlobalPose(
-    int trajectory_id) {
-    if (GetLocalSlamResultPoses().size() > 0) {
-        auto latest_local_pose_ = GetLocalSlamResultPoses().back();
-        auto local_transform =
-            map_builder_->pose_graph()->GetLocalToGlobalTransform(
-                trajectory_id);
-
-        return local_transform * latest_local_pose_;
-    } else {
-        return cartographer::transform::Rigid3d();
-    }
-}
-
 cartographer::transform::Rigid3d MapBuilder::GetGlobalPose(
-    int trajectory_id, cartographer::transform::Rigid3d& latest_local_pose_) {
+    int trajectory_id, cartographer::transform::Rigid3d& local_pose) {
     auto local_transform =
         map_builder_->pose_graph()->GetLocalToGlobalTransform(trajectory_id);
-    return local_transform * latest_local_pose_;
+    return local_transform * local_pose;
 }
 
 void MapBuilder::OverwriteOptimizeEveryNNodes(int value) {
--- a/slam-libraries/viam-cartographer/src/mapping/map_builder.cc
+++ b/slam-libraries/viam-cartographer/src/mapping/map_builder.cc
@@ -98,16 +98,16 @@ cartographer::sensor::TimedPointCloudData MapBuilder::GetDataFromFile(
 // Might be worth investigating in the future.
 cartographer::transform::Rigid3d MapBuilder::GetLastGlobalPose(
     int trajectory_id) {
-    if(GetLocalSlamResultPoses().size() > 0){
+    if (GetLocalSlamResultPoses().size() > 0) {
         auto latest_local_pose_ = GetLocalSlamResultPoses().back();
         auto local_transform =
-            map_builder_->pose_graph()->GetLocalToGlobalTransform(trajectory_id);
+            map_builder_->pose_graph()->GetLocalToGlobalTransform(
+                trajectory_id);
 
         return local_transform * latest_local_pose_;
-    }else{
+    } else {
         return cartographer::transform::Rigid3d();
     }
-    
 }
 
 cartographer::transform::Rigid3d MapBuilder::GetGlobalPose(
--- a/slam-libraries/viam-cartographer/src/mapping/map_builder.cc
+++ b/slam-libraries/viam-cartographer/src/mapping/map_builder.cc
@@ -105,6 +105,15 @@ cartographer::sensor::TimedPointCloudData MapBuilder::GetDataFromFile(
 
 // TODO: There might still be a lot of room to improve accuracy & speed.
 // Might be worth investigating in the future.
+cartographer::transform::Rigid3d MapBuilder::GetLastGlobalPose(
+    int trajectory_id) {
+    auto latest_local_pose_ = GetLocalSlamResultPoses().back();
+    auto local_transform =
+        map_builder_->pose_graph()->GetLocalToGlobalTransform(trajectory_id);
+
+    return local_transform * latest_local_pose_;
+}
+
 cartographer::transform::Rigid3d MapBuilder::GetGlobalPose(
     int trajectory_id, cartographer::transform::Rigid3d& latest_local_pose_) {
     auto local_transform =
--- a/slam-libraries/viam-cartographer/src/mapping/map_builder.cc
+++ b/slam-libraries/viam-cartographer/src/mapping/map_builder.cc
@@ -98,11 +98,16 @@ cartographer::sensor::TimedPointCloudData MapBuilder::GetDataFromFile(
 // Might be worth investigating in the future.
 cartographer::transform::Rigid3d MapBuilder::GetLastGlobalPose(
     int trajectory_id) {
-    auto latest_local_pose_ = GetLocalSlamResultPoses().back();
-    auto local_transform =
-        map_builder_->pose_graph()->GetLocalToGlobalTransform(trajectory_id);
+    if(GetLocalSlamResultPoses().size() > 0){
+        auto latest_local_pose_ = GetLocalSlamResultPoses().back();
+        auto local_transform =
+            map_builder_->pose_graph()->GetLocalToGlobalTransform(trajectory_id);
 
-    return local_transform * latest_local_pose_;
+        return local_transform * latest_local_pose_;
+    }else{
+        return cartographer::transform::Rigid3d();
+    }
+    
 }
 
 cartographer::transform::Rigid3d MapBuilder::GetGlobalPose(
--- a/slam-libraries/viam-cartographer/src/mapping/map_builder.h
+++ b/slam-libraries/viam-cartographer/src/mapping/map_builder.h
@@ -40,6 +40,9 @@ class MapBuilder {
         trajectory_builder_options_;
     std::vector<::cartographer::transform::Rigid3d> GetLocalSlamResultPoses();
 
+    cartographer::transform::Rigid3d GetLastGlobalPose(
+        int trajectory_id);
+
     cartographer::transform::Rigid3d GetGlobalPose(
         int trajectory_id,
         cartographer::transform::Rigid3d& latest_local_pose_);
--- a/slam-libraries/viam-cartographer/src/mapping/map_builder.h
+++ b/slam-libraries/viam-cartographer/src/mapping/map_builder.h
@@ -39,8 +39,7 @@ class MapBuilder {
         trajectory_builder_options_;
     std::vector<::cartographer::transform::Rigid3d> GetLocalSlamResultPoses();
 
-    cartographer::transform::Rigid3d GetLastGlobalPose(
-        int trajectory_id);
+    cartographer::transform::Rigid3d GetLastGlobalPose(int trajectory_id);
 
     cartographer::transform::Rigid3d GetGlobalPose(
         int trajectory_id,
--- a/slam-libraries/viam-cartographer/src/mapping/map_builder.h
+++ b/slam-libraries/viam-cartographer/src/mapping/map_builder.h
@@ -39,8 +39,6 @@ class MapBuilder {
         trajectory_builder_options_;
     std::vector<::cartographer::transform::Rigid3d> GetLocalSlamResultPoses();
 
-    cartographer::transform::Rigid3d GetLastGlobalPose(int trajectory_id);
-
     cartographer::transform::Rigid3d GetGlobalPose(
         int trajectory_id,
         cartographer::transform::Rigid3d& latest_local_pose_);
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -19,12 +19,9 @@ std::atomic<bool> b_continue_session{true};
 ::grpc::Status SLAMServiceImpl::GetPosition(ServerContext *context,
                                             const GetPositionRequest *request,
                                             GetPositionResponse *response) {
-    LOG(ERROR) << "GetPosition is not yet implemented.\n";
-    return grpc::Status(grpc::StatusCode::UNIMPLEMENTED,
-                        "GetPosition is not yet implemented.");
     cartographer::transform::Rigid3d global_pose;
     {
-        std::lock_guard<std::mutex> lk(map_builder_mutex);
+        std::lock_guard<std::mutex> lk(pose_mutex);
         global_pose = latest_global_pose;
     }
     // Setup mapping of pose message to the response. NOTE not using
@@ -53,6 +50,8 @@ std::atomic<bool> b_continue_session{true};
         global_pose.rotation().y());
     q->mutable_fields()->operator[]("kmag").set_number_value(
         global_pose.rotation().z());
+
+    return grpc::Status::OK;
 }
 
 ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
@@ -316,7 +315,7 @@ std::string SLAMServiceImpl::GetNextDataFile() {
 
 void SLAMServiceImpl::CreateMap() {
     cartographer::mapping::TrajectoryBuilderInterface *trajectory_builder;
-    int trajectory_id = 0;
+    int trajectory_id;
     {
         std::lock_guard<std::mutex> lk(map_builder_mutex);
         // Build TrajectoryBuilder
@@ -350,6 +349,7 @@ void SLAMServiceImpl::CreateMap() {
 
                 auto local_poses = map_builder.GetLocalSlamResultPoses();
                 if (local_poses.size() > 0) {
+                    std::lock_guard<std::mutex> lk(pose_mutex);
                     latest_global_pose = map_builder.GetGlobalPose(
                         trajectory_id, local_poses.back());
                 }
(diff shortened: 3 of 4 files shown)


slam-libraries/viam-cartographer/src/mapping/map_builder.cc
```diff
@@ -96,6 +96,20 @@ cartographer::sensor::TimedPointCloudData MapBuilder::GetDataFromFile(
 
 // TODO: There might still be a lot of room to improve accuracy & speed.
 // Might be worth investigating in the future.
+cartographer::transform::Rigid3d MapBuilder::GetLastGlobalPose(
```
kkufieta, line 99: `MapBuilder::GetLatestGlobalPose`

slam-libraries/viam-cartographer/src/mapping/map_builder.cc
```diff
@@ -96,6 +96,20 @@ cartographer::sensor::TimedPointCloudData MapBuilder::GetDataFromFile(
...
+    int trajectory_id) {
+    if (GetLocalSlamResultPoses().size() > 0) {
```
kkufieta, line 101: Pull out `GetLocalSlamResultPoses()` and save the result in `local_poses`. Reuse `local_poses`, that way you don't have to call `GetLocalSlamResultPoses()` twice (which might be doing a lot of work behind the scenes)

slam-libraries/viam-cartographer/src/mapping/map_builder.cc
```diff
@@ -96,6 +96,20 @@ cartographer::sensor::TimedPointCloudData MapBuilder::GetDataFromFile(
...
+        auto latest_local_pose_ = GetLocalSlamResultPoses().back();
```
kkufieta, line 102: [nit] Rename `latest_local_pose_` to `latest_local_pose`. The trailing `_` is just an artifact

slam-libraries/viam-cartographer/src/mapping/map_builder.cc
```diff
@@ -96,6 +96,20 @@ cartographer::sensor::TimedPointCloudData MapBuilder::GetDataFromFile(
...
+        auto local_transform =
+            map_builder_->pose_graph()->GetLocalToGlobalTransform(
+                trajectory_id);
+
+        return local_transform * latest_local_pose_;
```
kkufieta, line 107: Call `GetGlobalPose` here to do this

slam-libraries/viam-cartographer/src/mapping/map_builder.cc
```diff
@@ -96,6 +96,20 @@ cartographer::sensor::TimedPointCloudData MapBuilder::GetDataFromFile(
...
+    } else {
+        return cartographer::transform::Rigid3d();
+    }
```
kkufieta, line 110: We shouldn't be returning the identity `Rigid3d` matrix here. This is the starting point of the robot within the map. If we for some reason wouldn't be able to find & return the most recent global pose, instead of setting it to (0,0,0) we should keep it at the latest value we found the robot to be. I suggest we raise an error here that indicates that we didn't find the most recent global pose.

slam-libraries/viam-cartographer/src/mapping/map_builder.h
```diff
@@ -39,6 +39,8 @@ class MapBuilder {
         trajectory_builder_options_;
     std::vector<::cartographer::transform::Rigid3d> GetLocalSlamResultPoses();
 
+    cartographer::transform::Rigid3d GetLastGlobalPose(int trajectory_id);
```
kkufieta, line 42: `GetLatestGlobalPose`

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,9 +19,45 @@ std::atomic<bool> b_continue_session{true};
...
                                             const GetPositionRequest *request,
                                             GetPositionResponse *response) {
-    LOG(ERROR) << "GetPosition is not yet implemented.\n";
-    return grpc::Status(grpc::StatusCode::UNIMPLEMENTED,
-                        "GetPosition is not yet implemented.");
+
+    cartographer::transform::Rigid3d global_pose;
+    {
+        std::lock_guard<std::mutex> lk(map_builder_mutex);
+        if (map_builder.GetLocalSlamResultPoses().size() > 0) {
```
kkufieta, line 26: I suggest we save the latest global pose as a member variable within the slam service and update it as we're building the map continuously. Then, we can use that member variable here.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,9 +19,45 @@ std::atomic<bool> b_continue_session{true};
...
+            global_pose = map_builder.GetLastGlobalPose(trajectory_id);
+            LOG(INFO) << "global_pose: " << global_pose.DebugString();
+        } else {
+            LOG(ERROR) << "GetPosition is not ready.\n";
+            return grpc::Status(grpc::StatusCode::UNAVAILABLE,
+                                "GetPosition is not ready yet");
+        }
+    }
+    // Setup mapping of pose message to the response. NOTE not using
+    // inFrame->set_reference_frame yet
```
kkufieta, line 36: [q] When are we going to use it? Is this going to be changed in a future ticket?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -285,7 +321,7 @@ std::string SLAMServiceImpl::GetNextDataFile() {
 
 void SLAMServiceImpl::CreateMap() {
     cartographer::mapping::TrajectoryBuilderInterface *trajectory_builder;
-    int trajectory_id;
```
kkufieta, line 288: I think we should keep this here, and for the global latest pose to be updated within this function. The reason for not making `trajectory_id` one global variable within the slam service is that `trajectory_id` is used ubiquitously everywhere within cartographer. I am not sure that we won't be juggling multiple `trajectory_ids` in the future, which is why I'd like for them to be kept as a local variable within functions for now.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -313,34 +346,18 @@ void SLAMServiceImpl::CreateMap() {
...
-
-                auto local_poses = map_builder.GetLocalSlamResultPoses();
-                if (local_poses.size() > 0) {
-                    auto latest_local_pose = local_poses.back();
-                    cartographer::transform::Rigid3d global_pose =
-                        map_builder.GetGlobalPose(trajectory_id,
-                                                  latest_local_pose);
-                    myfile << "global_pose: " << global_pose.DebugString()
-                           << std::endl;
-                }
```
kkufieta, line 335: Let's set `latest_global_pose` as a member variable within the slam service that's continuously updated here every time we're updating the map. We can then use that global pose when we want to read & return it in `GetPosition`

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -131,6 +135,7 @@ class SLAMServiceImpl final : public SLAMService::Service {
     double occupied_space_weight = 20.0;
     double translation_weight = 10.0;
     double rotation_weight = 1.0;
+    int trajectory_id = 0;
```
kkufieta, line 138: Let's not make `trajectory_id` a "global" slam service variable for now. I am not sure if we won't be handling multiple trajectory ids in other slam service functions in the future, so we should keep this variable local to functions as long as we can do so.

kkufieta (changes requested):

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: Kk so make pose global and trajectory_id local

slam-libraries/viam-cartographer/src/mapping/map_builder.cc
JohnN193: The return here is to ensure something is set in the function based off what the function has access to. The corresponding check in the slam service prevents the function from running if no local poses exist

JohnN193: I can add a log for the error

tessavitabile (commented):
No comments beyond Kat's!

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: I agree here. This is what we do for orbslam as well. We keep the map and position as member variables and update them as part of mapping. That way calls to GetMap and GetPosition can use the previous map/position if mapping is in progress.

slam-libraries/viam-cartographer/src/mapping/map_builder.cc
kkufieta: I don't think that's a good idea, because it can easily become a source of error to return a value that does not really exist. Meaning in this case, if there is no value to be found, then returning an identity value is incorrect, since there is technically no value to be returned. 

I would either raise an error if it can't find the latest pose or remove this function and leave the code where it originally was in the `CreateMap` function - which would solve this dilemma.

JohnN193: kk I'll revert back

JohnN193:
Based off suggestions I removed the additional function and just used the `GetGlobalPose` function

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: [q] Curious if this comment should have an associated ticket with it, so we can provide more information and/or know when this will be changed?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -285,7 +316,7 @@ std::string SLAMServiceImpl::GetNextDataFile() {
 
 void SLAMServiceImpl::CreateMap() {
     cartographer::mapping::TrajectoryBuilderInterface *trajectory_builder;
-    int trajectory_id;
+    int trajectory_id = 0;
```
kkufieta, line 319: [nit] I think remove the initialization with a value of 0? My guess is that this way we'll have to overwrite it somewhere with a meaningful value - else it'll throw an error when we try to use it? 

Don't want `trajectory_id` to be set to a value that's wrong (e.g. a value of `0`) and use that accidentally. Not sure about this, curious what you think @tessavitabile .

kkufieta (commented):
Looks good - just a question and nit (with questionmark)

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -22,6 +22,37 @@ ::grpc::Status SLAMServiceImpl::GetPosition(ServerContext *context,
     LOG(ERROR) << "GetPosition is not yet implemented.\n";
     return grpc::Status(grpc::StatusCode::UNIMPLEMENTED,
                         "GetPosition is not yet implemented.");
+    cartographer::transform::Rigid3d global_pose;
+    {
+        std::lock_guard<std::mutex> lk(map_builder_mutex);
```
kkufieta, line 27: Actually this should not be locking the `map_builder_mutex`. We do instead probably need a `slam_mutex` that'll lock when we change/get the global pose.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -22,6 +22,37 @@ ::grpc::Status SLAMServiceImpl::GetPosition(ServerContext *context,
...
+        global_pose = latest_global_pose;
```
kkufieta, line 28: [q] Once we copy `latest_global_pose` into `global_pose`, would changing `latest_global_pose` also change `global_pose`? I want to check whether or not it is safe to use `global_pose` after releasing the mutex.

kkufieta (changes requested):
Found something I missed last time + one more question.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: Currently we do not plan to use the reference frame part of our pose message. When we redo our [proto methods](https://viam.atlassian.net/browse/DATA-431) in the map representation tickets the reference frame of the proto will be phased out. I think we should still note the reference frame is not being used(can remove the yet)

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: I believe it should not change, we are just assigning the values directly here

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: The first thing that happens in CreateMap is setting `trajectory_id` so I am not too worried about using 0. Also I like setting an initialized value so its not using a random number if something bad happens

tessavitabile: In C++, writing `int trajectory_id;` and `int trajectory_id = 0;` is the same, since 0 is the default value for `int`. I'm also not worried about failing to set this value, since we set it immediately below.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: Agreed. Also, would it be better to store the trajectory nodes as a member variable and update them when we update `latest_global_pose`?  This would ensure that `GetMap` doesn't need to block if cartographer is currently processing data. Should we file a ticket for this? I wish I had thought of this during the review for https://viam.atlassian.net/browse/DATA-147 or https://viam.atlassian.net/browse/DATA-159. This may also be part of why it takes a long time to paint a jpeg map (https://viam.atlassian.net/browse/DATA-661).

tessavitabile: I will file a ticket for this.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I think for this I'd probably check what happens if you copy `Eigen::Matrix<FloatType, 3, 1>` and `Eigen::Quaternion<FloatType>`, since those are underlying the `Rigid3d` data structure from what I can see. Question is if those are safe to copy & access later without a mutex

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: https://viam.atlassian.net/browse/DATA-706

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: Verified this was safe with kat in person.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -143,6 +147,10 @@ class SLAMServiceImpl final : public SLAMService::Service {
 
     std::mutex map_builder_mutex;
     mapping::MapBuilder map_builder;
+
+    std::mutex pose_mutex;
```
tessavitabile, line 151: [opt] We might want to call this `response_mutex` or something else if we add `trajectory_nodes` here, but that can also wait until https://viam.atlassian.net/browse/DATA-706

kkufieta (approved):
LGTM

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -313,34 +340,25 @@ void SLAMServiceImpl::CreateMap() {
...
 
                 auto local_poses = map_builder.GetLocalSlamResultPoses();
                 if (local_poses.size() > 0) {
-                    auto latest_local_pose = local_poses.back();
-                    cartographer::transform::Rigid3d global_pose =
-                        map_builder.GetGlobalPose(trajectory_id,
-                                                  latest_local_pose);
-                    myfile << "global_pose: " << global_pose.DebugString()
-                           << std::endl;
+                    std::lock_guard<std::mutex> lk(pose_mutex);
```
tessavitabile, line 352: [nit] It would be better to avoid taking `pose_mutex` while holding `map_builder_mutex`. Otherwise, we'll always need to impose the ordering that if a codepath is going to take both mutexes, it must take `map_builder_mutex` first. You can do this by copying the pose into a local variable (while holding only `map_builder_mutex`), then after `map_builder_mutex` is released, you can lock `pose_mutex` and copy the local variable into `latest_global_pose`.

If that seems like too much copying, it's okay to leave this as is, but then please add a comment in the header file that any codepath that locks both `map_builder_mutex` and `pose_mutex` must lock `map_builder_mutex` first.

tessavitabile (approved):
LGTM mod nit

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
kkufieta: +1 on this, either now or in the future. I also like `slam_mutex` but might be not descriptive enough

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: +1 on copying rather than imposing an order

JohnN193: yeah wasn't sure on doing the double mutex there. I'll move the grpc mutex out.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -143,6 +147,10 @@ class SLAMServiceImpl final : public SLAMService::Service {
...
+    cartographer::transform::Rigid3d latest_global_pose =
```
jeremyhyde-viam, line 152: [q] why isn't this just global pose? Is there another variable with that name already im not seeing?

jeremyhyde-viam (approved):
LGTM after an initial pass. Still getting caught up on cartographer improvements. I just have minor questions regarding code reusability and a variable name.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
kkufieta: I think either is fine. `latest_global_pose` feels more accurate since we can not guarantee to know the `global_pose` at all times, only the latest global pose we grabbed & updated. Adding `latest_` therefore just implies that there's a certain amount of uncertainty in this value. But I don't think it matters much either way.

JohnN193: yeah latest just kinda stuck. Wanted it to be distinct because we were grabbing a specific point in our trajectory

jeremyhyde-viam: gotcha, was just confused about the switch in local pose to remove latest and why it was kept here

JohnN193: oh yeah the function doesn't necessarily need the latest pose which is why we removed it from that

PR #100, opened 2022-11-01
[DATA-568] Add cartographer integration tests

tessavitabile:
https://viam.atlassian.net/browse/DATA-568

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -16,6 +16,6 @@ on:
 
 jobs:
   test:
-    uses: viamrobotics/slam/.github/workflows/test.yml@main
+    uses: viamrobotics/slam/.github/workflows/test.yml@DATA-568
     secrets:
       ARTIFACT_READ_ONLY_GCP_CREDENTIALS: ${{ secrets.ARTIFACT_READ_ONLY_GCP_CREDENTIALS }}
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -316,7 +316,7 @@ std::string SLAMServiceImpl::GetNextDataFileOffline() {
         throw std::runtime_error("no data in data directory");
     }
     if (current_file_offline == file_list_offline.size()) {
-        LOG(INFO) << "Finished processing offline images";
+        LOG(INFO) << "Finished processing offline data";
         return "";
     }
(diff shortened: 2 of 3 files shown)


.github/workflows/pullrequest.yml
```diff
@@ -16,6 +16,6 @@ on:
 
 jobs:
   test:
-    uses: viamrobotics/slam/.github/workflows/test.yml@main
+    uses: viamrobotics/slam/.github/workflows/test.yml@DATA-568
```
tessavitabile, line 19: Will revert before merging

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -316,7 +316,7 @@ std::string SLAMServiceImpl::GetNextDataFileOffline() {
         throw std::runtime_error("no data in data directory");
     }
     if (current_file_offline == file_list_offline.size()) {
-        LOG(INFO) << "Finished processing offline images";
+        LOG(INFO) << "Finished processing offline data";
```
tessavitabile, line 319: I thought it would be more correct to say "data" than "images" for cartographer.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -403,6 +403,7 @@ void SLAMServiceImpl::CreateMap() {
                 }
             }
         }
+        VLOG(1) << "Passed sensor data to SLAM " << file;
```
tessavitabile, line 406: This log line is needed by the integration tests.

kkufieta: [nit] I think it would be helpful to add a comment that clarifies that this log is needed for integration tests, such that someone doesn't accidentally delete it.

[q] Why do you use `VLOG(1)` here when we usually use `LOG(INFO)` in cartographer?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: [nit] I think it would be helpful to add a comment that clarifies that this log is needed for integration tests, such that someone doesn't accidentally delete it

kkufieta (commented):
A couple of nits and a question

kkufieta (approved):
LGTM! Thanks for protecting all the important log lines with comments :)

JohnN193 (approved):
LGTM

PR #101, opened 2022-11-02
[DATA-114] Add apriori map loading

kkufieta:
https://viam.atlassian.net/browse/DATA-114

Add apriori map loading.


slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -407,6 +431,10 @@ void SLAMServiceImpl::CreateMap() {
     cartographer::transform::Rigid3d tmp_global_pose =
         cartographer::transform::Rigid3d();
     while (file != "") {
+        if (file.find(".pcd") == std::string::npos) {
+            file = GetNextDataFile();
+            continue;
+        }
```
kkufieta, line 437: This ensures that if there's e.g. a DS_Store file present, that we're not trying to read it in.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -319,28 +329,47 @@ bool SLAMServiceImpl::ExtractPointCloudToBuffer(std::stringbuf &buffer) {
...
+    if (action_mode == SLAMServiceActionMode::UPDATING ||
+        action_mode == SLAMServiceActionMode::LOCALIZING) {
+        // Check if there is an apriori map in the path_to_map directory
+        std::vector<std::string> map_filenames =
+            viam::io::ListSortedFilesInDirectory(path_to_map);
+
+        bool found_map = false;
+        std::string latest_map_filename;
+        for (size_t i = map_filenames.size() - 1; i == 0; i--) {
+            if (map_filenames.at(i).find(".pbstream") != std::string::npos) {
```
kkufieta, line 343: This ensures that if there's e.g. a DS_Store file present, that we're not counting it or trying to read it in.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -105,22 +105,32 @@ ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
...
-        LOG(INFO) << "Running in localization only mode";
-        return SLAMServiceActionMode::LOCALIZING;
+void SLAMServiceImpl::DetermineActionMode() {
+    // Check if there is an apriori map in the path_to_map directory
+    std::vector<std::string> map_filenames =
+        viam::io::ListSortedFilesInDirectory(path_to_map);
+
+    // Check if there is a *.pbstream map in the path_to_map directory
+    for (auto filename : map_filenames) {
+        if (filename.find(".pbstream") != std::string::npos) {
```
jeremyhyde-viam, line 115: [opt] potential split of filtering algorithm to take in a list of strings, a substring and return list of all matching ones. This can then be checked for first, last or any occurrences easily (maybe reused in data filtering as well)

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -105,22 +105,32 @@ ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
...
-    if (map_rate_sec.count() == -1) {
-        LOG(INFO) << "Running in localization only mode";
-        return SLAMServiceActionMode::LOCALIZING;
+void SLAMServiceImpl::DetermineActionMode() {
+    // Check if there is an apriori map in the path_to_map directory
+    std::vector<std::string> map_filenames =
+        viam::io::ListSortedFilesInDirectory(path_to_map);
+
+    // Check if there is a *.pbstream map in the path_to_map directory
+    for (auto filename : map_filenames) {
```
jeremyhyde-viam, line 114: [q] the slam directory is very heavily controlled, can we assume that any file in the map directory must be a pbstream?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -105,22 +105,32 @@ ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
...
-    // map is detected and loaded. Will be implemented in this ticket:
-    // https://viam.atlassian.net/browse/DATA-114
-
-    // TODO: Change this depending on the outcome of this scope
-    // doc:
-    // https://docs.google.com/document/d/1RsT-c0QOtkMKa-rwUGY0-emUmKIRSaO3mzT1v6MtFjk/edit#heading=h.tcicyojyqi6c
-    if (map_rate_sec.count() == -1) {
-        LOG(INFO) << "Running in localization only mode";
-        return SLAMServiceActionMode::LOCALIZING;
+void SLAMServiceImpl::DetermineActionMode() {
```
jeremyhyde-viam, line 108: [opt] SetActionMode (consistency with GetActionMode)

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -319,28 +329,47 @@ bool SLAMServiceImpl::ExtractPointCloudToBuffer(std::stringbuf &buffer) {
 void SLAMServiceImpl::ProcessData() {
     // Set up and build the MapBuilder
     SetUpMapBuilder();
+    DetermineActionMode();
+
+    if (action_mode == SLAMServiceActionMode::UPDATING ||
```
jeremyhyde-viam, line 334: [nit] use GetActionMode or remove function if unused

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -319,28 +329,47 @@ bool SLAMServiceImpl::ExtractPointCloudToBuffer(std::stringbuf &buffer) {
 void SLAMServiceImpl::ProcessData() {
     // Set up and build the MapBuilder
     SetUpMapBuilder();
+    DetermineActionMode();
```
jeremyhyde-viam, line 332: [q] should DetemrineActionMode return the map filename if one is located? Thus removing need for second for loop below?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -319,28 +329,47 @@ bool SLAMServiceImpl::ExtractPointCloudToBuffer(std::stringbuf &buffer) {
...
+
+    if (action_mode == SLAMServiceActionMode::UPDATING ||
+        action_mode == SLAMServiceActionMode::LOCALIZING) {
+        // Check if there is an apriori map in the path_to_map directory
+        std::vector<std::string> map_filenames =
+            viam::io::ListSortedFilesInDirectory(path_to_map);
+
+        bool found_map = false;
+        std::string latest_map_filename;
+        for (size_t i = map_filenames.size() - 1; i == 0; i--) {
```
jeremyhyde-viam, line 342: see previous comments

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -319,28 +329,47 @@ bool SLAMServiceImpl::ExtractPointCloudToBuffer(std::stringbuf &buffer) {
...
+                latest_map_filename = map_filenames.at(i);
+                found_map = true;
+                break;
+            }
+        }
+        if (!found_map) {
+            throw std::runtime_error(
+                "cannot find maps but they should be present");
+        }
+        bool optimize = true;
```
jeremyhyde-viam, line 353: [major] optimize is supposed to be passed in as a config_param. This takes time and there are instances where someone may have already optimized a map on a previous run and so can skip this step

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -319,28 +329,47 @@ bool SLAMServiceImpl::ExtractPointCloudToBuffer(std::stringbuf &buffer) {
...
+        bool load_frozen_trajectory = true;  // true for LOCALIZING action mode
```
jeremyhyde-viam, line 354: `bool load_frozen_trajectory = action_mode == SLAMServiceActionMode::LOCALIZING`

jeremyhyde-viam: Note: maybe dont need to even create the variable if the check can be done inside `LoadFromFile`

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -382,20 +411,15 @@ std::string SLAMServiceImpl::GetNextDataFile() {
...
-            map_builder.GetLocalSlamResultCallback());
-
-        LOG(INFO) << "Trajectory ID: " << trajectory_id;
-
-        trajectory_builder =
-            map_builder.map_builder_->GetTrajectoryBuilder(trajectory_id);
+        // Set TrajectoryBuilder
+        trajectory_id = map_builder.SetTrajectoryBuilder(&trajectory_builder,
+                                                         {kRangeSensorId});
+        LOG(INFO) << "Using trajectory ID: " << trajectory_id;
```
jeremyhyde-viam, line 422: [opt]  `Using new trajectory ID:` (if comment from early is taken to add from apriori map to log of other trajectory ids)

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -407,6 +431,10 @@ void SLAMServiceImpl::CreateMap() {
     cartographer::transform::Rigid3d tmp_global_pose =
         cartographer::transform::Rigid3d();
     while (file != "") {
```
jeremyhyde-viam, line 433: Consider

```
string file;
while (( file = GetNextDataFile()) != "") {
    ... do thing...
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
jeremyhyde-viam: This check should be included in GetNextDataFile

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -73,8 +73,13 @@ class SLAMServiceImpl final : public SLAMService::Service {
     // empty string if stop has been signaled.
     std::string GetNextDataFileOnline();
 
-    // CreateMap creates a map from scratch.
-    void CreateMap();
+    // RunSLAM runs SLAM in the SLAMServiceActionMode mode: Either mapping from
+    // scratch, updating an apriori map, or localizing on an apriori map.
```
jeremyhyde-viam, line 77: [opt] instead of `scratch` say `creating a new map`

jeremyhyde-viam (changes requested):
Generally looks good! I love the splitting off on functions for apriori map loading as well as the action types. Some suggestions for optimization and making code more succinct but  most are opt or minor changes

slam-libraries/viam-cartographer/src/mapping/map_builder.cc
```diff
@@ -57,6 +57,30 @@ void MapBuilder::BuildMapBuilder() {
...
+void MapBuilder::LoadFromFile(std::string map_filename,
+                              bool load_frozen_trajectory, bool optimize) {
+    std::map<int, int> trajectory_ids_map =
+        map_builder_->LoadStateFromFile(map_filename, load_frozen_trajectory);
+
+    if (optimize) {
+        map_builder_->pose_graph()->RunFinalOptimization();
+    }
+    for (auto&& trajectory_ids_pair : trajectory_ids_map)
+        LOG(INFO) << "Trajectory ids mapping: " << trajectory_ids_pair.first
```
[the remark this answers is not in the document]
kkufieta: We're not using BOOST because cartographer uses glog and this way we can have one standardized logging framework to forward all logs for viam-cartographer.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: We can not, especially because cartographer runs on osx, as soon as a user takes a look into the folder with data a `.DS_Store` is generated and our code fails.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: That's right, and as soon as we implement the part where we pass in `optimize` via the `config_param`, we can use this here to check whether or not we're going to optimize it. For now it's just a variable set in the code - happy to set it to false if we prefer that. But this implements the framework for letting the user change this flag.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -105,22 +105,32 @@ ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
...
+    // Check if there is an apriori map in the path_to_map directory
+    std::vector<std::string> map_filenames =
+        viam::io::ListSortedFilesInDirectory(path_to_map);
+
+    // Check if there is a *.pbstream map in the path_to_map directory
+    for (auto filename : map_filenames) {
+        if (filename.find(".pbstream") != std::string::npos) {
+            // There is an apriori map present, so we're running either in
+            // updating or localization mode.
+            if (map_rate_sec.count() == 0) {
```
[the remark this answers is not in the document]
kkufieta: This check is only executed once after a `.pbstream` file is found. Once that file is found, the function returns after setting the `action_mode` either to localizing or updating.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I don't think it should return a map filename, since its job is merely to determine the action mode. I was also thinking about how to (or whether or not) to remove the second time we're looping over the map folder, but decided that I want to separate finding the action mode from finding the latest map filename from each other, and running the loop twice is not going to be a big overhead. So I chose modularity over performance improvement.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: `GetActionMode` is used in a few places outside of `slam_service.cc` (in tests and in `config.cc`). I think since this is implemented within the slam service, we can use the member variable directly instead of calling the function and saving the variable in a new variable name. I went through slam_service to make sure we're not using `GetActionMode` within its functions (doing that was an artifact from how it was implemented before).

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: Added a comment above the flag pointing to the ticket where it's going to be implemented at.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: Do you mean that checking if the file is a pcd file or not should be implemented within `GetNextDataFile`?

jeremyhyde-viam: I personally think so since it will have to be run anytime `GetNextDataFile` is run and we are running additional checks there, nice to have them all in one space.

slam-libraries/viam-cartographer/src/mapping/map_builder.cc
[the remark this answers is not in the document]
jeremyhyde-viam: Cool! Thanks for the clairification!

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
jeremyhyde-viam: can we just force a deletion of .DS_Store if detected?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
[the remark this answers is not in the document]
jeremyhyde-viam: Good point, I'd still prefer to have it outside the loop for clarity i.e.

```
Loop:
 - check if valid map exists
 
if valid map exists && map_rate_sec.count() == 0
- action = localize
else if valid map exists && map_rate_sec.count() != 0
- action = update
else 
- action = mapping 
```

however thats just more easily legible to me. Plus you could add in an error for if no map exists and map_rate_sec.count() == 0, if we haven't done that already somewhere else

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
jeremyhyde-viam: got it, would be nice to not have to loop multiple times but due to the low number of expected maps in the folder and for the sake of code clarity lets keep it as is

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
jeremyhyde-viam: Oh darn, I github's search was not returning any results for it so I checked in main.cc and map_builder.cc and didnt see it being used. Didnt check config

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
jeremyhyde-viam: Perfect :)

jeremyhyde-viam (approved):
A few minor potnetial edits but otherwise LGTM

PR #103, opened 2022-11-03
[Data-657] carto implement pose marker

JohnN193:
Implement an optional marker for the robot's current location on the jpeg map endpoint
Screen Shot 2022-11-03 at 4 58 18 PM

https://viam.atlassian.net/browse/DATA-657

### DIFF
--- a/slam-libraries/viam-cartographer/src/io/submap_painter.cc
+++ b/slam-libraries/viam-cartographer/src/io/submap_painter.cc
@@ -106,7 +106,7 @@ cartographer::io::PaintSubmapSlicesResult PaintSubmapSlices(
                 cairo_set_source_surface(cr.get(), submap_slice.surface.get(),
                                          0., 0.);
                 cairo_paint(cr.get());
-            });        
+            });
         cairo_surface_flush(surface.get());
     }
     return cartographer::io::PaintSubmapSlicesResult(std::move(surface),
@@ -114,17 +114,23 @@ cartographer::io::PaintSubmapSlicesResult PaintSubmapSlices(
 }
 
 // Draw a position on the current map
-void DrawPositionOnSurface(cartographer::io::PaintSubmapSlicesResult *painted_slices, cartographer::transform::Rigid3d global_pose, float resolution){
+void DrawPositionOnSurface(
+    cartographer::io::PaintSubmapSlicesResult* painted_slices,
+    cartographer::transform::Rigid3d global_pose, float resolution) {
     float marker_size = .3;
-    
-    auto cr = cartographer::io::MakeUniqueCairoPtr(cairo_create(painted_slices->surface.get()));
+
+    auto cr = cartographer::io::MakeUniqueCairoPtr(
+        cairo_create(painted_slices->surface.get()));
 
     // move to origin of map
-    cairo_translate(cr.get(), painted_slices->origin.x(), painted_slices->origin.y());
-    cairo_set_source_rgba(cr.get(),1,0,0,1);
+    cairo_translate(cr.get(), painted_slices->origin.x(),
+                    painted_slices->origin.y());
+    cairo_set_source_rgba(cr.get(), 1, 0, 0, 1);
 
-    //draw circle at desired global pose
-    cairo_arc(cr.get(), global_pose.translation().x() / resolution, -global_pose.translation().y() / resolution, marker_size / resolution, 0, 2*M_PI);
+    // draw circle at desired global pose
+    cairo_arc(cr.get(), global_pose.translation().x() / resolution,
+              -global_pose.translation().y() / resolution,
+              marker_size / resolution, 0, 2 * M_PI);
     cairo_fill(cr.get());
     cairo_surface_flush(painted_slices->surface.get());
 }
--- a/slam-libraries/viam-cartographer/src/io/submap_painter.cc
+++ b/slam-libraries/viam-cartographer/src/io/submap_painter.cc
@@ -106,11 +106,27 @@ cartographer::io::PaintSubmapSlicesResult PaintSubmapSlices(
                 cairo_set_source_surface(cr.get(), submap_slice.surface.get(),
                                          0., 0.);
                 cairo_paint(cr.get());
-            });
+            });        
         cairo_surface_flush(surface.get());
     }
     return cartographer::io::PaintSubmapSlicesResult(std::move(surface),
                                                      origin);
+}
+
+// Draw a position on the current map
+void DrawPositionOnSurface(cartographer::io::PaintSubmapSlicesResult *painted_slices, cartographer::transform::Rigid3d global_pose, float resolution){
+    float marker_size = .3;
+    
+    auto cr = cartographer::io::MakeUniqueCairoPtr(cairo_create(painted_slices->surface.get()));
+
+    // move to origin of map
+    cairo_translate(cr.get(), painted_slices->origin.x(), painted_slices->origin.y());
+    cairo_set_source_rgba(cr.get(),1,0,0,1);
+
+    //draw circle at desired global pose
+    cairo_arc(cr.get(), global_pose.translation().x() / resolution, -global_pose.translation().y() / resolution, marker_size / resolution, 0, 2*M_PI);
+    cairo_fill(cr.get());
+    cairo_surface_flush(painted_slices->surface.get());
 }
 
 }  // namespace io
--- a/slam-libraries/viam-cartographer/src/io/submap_painter.cc
+++ b/slam-libraries/viam-cartographer/src/io/submap_painter.cc
@@ -113,8 +113,8 @@ cartographer::io::PaintSubmapSlicesResult PaintSubmapSlices(
                                                      origin);
 }
 
-// Draw a position on the current map
-void DrawPositionOnSurface(
+// Draw a marker for the current pose within the map
+void DrawPoseOnSurface(
     cartographer::io::PaintSubmapSlicesResult* painted_slices,
     cartographer::transform::Rigid3d global_pose, float resolution) {
     float marker_size = .3;
--- a/slam-libraries/viam-cartographer/src/io/submap_painter.h
+++ b/slam-libraries/viam-cartographer/src/io/submap_painter.h
@@ -21,7 +21,7 @@ cartographer::io::PaintSubmapSlicesResult PaintSubmapSlices(
                    cartographer::io::SubmapSlice>& submaps,
     double resolution);
 
-void DrawPositionOnSurface(
+void DrawPoseOnSurface(
     cartographer::io::PaintSubmapSlicesResult* painted_slices,
     cartographer::transform::Rigid3d global_pose, float resolution);
 
--- a/slam-libraries/viam-cartographer/src/io/submap_painter.h
+++ b/slam-libraries/viam-cartographer/src/io/submap_painter.h
@@ -21,7 +21,9 @@ cartographer::io::PaintSubmapSlicesResult PaintSubmapSlices(
                    cartographer::io::SubmapSlice>& submaps,
     double resolution);
 
-void DrawPositionOnSurface(cartographer::io::PaintSubmapSlicesResult *painted_slices, cartographer::transform::Rigid3d global_pose, float resolution);
+void DrawPositionOnSurface(
+    cartographer::io::PaintSubmapSlicesResult* painted_slices,
+    cartographer::transform::Rigid3d global_pose, float resolution);
 
 }  // namespace io
 }  // namespace viam
--- a/slam-libraries/viam-cartographer/src/io/submap_painter.h
+++ b/slam-libraries/viam-cartographer/src/io/submap_painter.h
@@ -21,6 +21,8 @@ cartographer::io::PaintSubmapSlicesResult PaintSubmapSlices(
                    cartographer::io::SubmapSlice>& submaps,
     double resolution);
 
+void DrawPositionOnSurface(cartographer::io::PaintSubmapSlicesResult *painted_slices, cartographer::transform::Rigid3d global_pose, float resolution);
+
 }  // namespace io
 }  // namespace viam
 
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -261,8 +261,7 @@ std::string SLAMServiceImpl::PaintMap(bool pose_marker_flag) {
             std::lock_guard<std::mutex> lk(viam_response_mutex);
             global_pose = latest_global_pose;
         }
-        viam::io::DrawPoseOnSurface(&painted_slices, global_pose,
-                                        kPixelSize);
+        viam::io::DrawPoseOnSurface(&painted_slices, global_pose, kPixelSize);
     }
 
     auto image = viam::io::Image(std::move(painted_slices.surface));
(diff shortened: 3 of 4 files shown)


slam-libraries/viam-cartographer/src/io/submap_painter.cc
```diff
@@ -113,5 +113,27 @@ cartographer::io::PaintSubmapSlicesResult PaintSubmapSlices(
                                                      origin);
 }
 
+// Draw a position on the current map
+void DrawPositionOnSurface(
```
JohnN193, line 117: added a new function here to mark the pose on top of the final map, that way we do not run into issues with masking & positions becoming faded

slam-libraries/viam-cartographer/src/io/submap_painter.cc
```diff
@@ -113,5 +113,27 @@ cartographer::io::PaintSubmapSlicesResult PaintSubmapSlices(
...
+    cartographer::io::PaintSubmapSlicesResult* painted_slices,
+    cartographer::transform::Rigid3d global_pose, float resolution) {
+    float marker_size = .3;
+
+    auto cr = cartographer::io::MakeUniqueCairoPtr(
+        cairo_create(painted_slices->surface.get()));
+
+    // move to origin of map
+    cairo_translate(cr.get(), painted_slices->origin.x(),
```
JohnN193, line 126: unsure of how this origin interacts with multiple trajectories/submaps in a pbstream. May have to test once apriori map PR goes through

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -449,6 +459,16 @@ void SLAMServiceImpl::CreateMap() {
 
         map_builder.map_builder_->FinishTrajectory(trajectory_id);
         map_builder.map_builder_->pose_graph()->RunFinalOptimization();
+        auto local_poses = map_builder.GetLocalSlamResultPoses();
```
JohnN193, line 462: initially pose was not being updated after an optimization occurs. Currently added this here for offline mode but not sure if there is a better place to include this

slam-libraries/viam-cartographer/src/io/submap_painter.cc
kkufieta: I think we should test this with apriori maps and localization before merging. Both will probably be done next week Monday.

kkufieta (changes requested):
Needs to be tested for updating and localizing as well.

JohnN193:
Testing with apriori map loading. Using the [feb_11_large](https://drive.google.com/drive/u/0/folders/1oIjVyhsXb6kvusrVqc7Bw4dKk_6PEw-s) dataset and updating/localizing with [feb_11_small](https://drive.google.com/drive/u/0/folders/164VJicRLSWqZcf-bxL9ij7K3JM0RPKll). 

initial map(only feb_11_large)
Screen Shot 2022-11-08 at 12 16 31 PM

using updating mode
Screen Shot 2022-11-08 at 12 23 19 PM

using localization mode
Screen Shot 2022-11-08 at 12 31 14 PM

kkufieta:
> fine with removing the original trajectory method with this PR. The submap generator still relies on the surfaces from that function so would need to pull those out

@JohnN193 I don't think this is worth the effort right now, since we can't really see the trajectory anyways, and as much as the jpeg image improvements are not going to be pursued in favor of implementing the generalized map interface, I think that removing the faded trajectories is equally not worth the time & effort

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -477,6 +487,16 @@ void SLAMServiceImpl::RunSLAM() {
 
         map_builder.map_builder_->FinishTrajectory(trajectory_id);
         map_builder.map_builder_->pose_graph()->RunFinalOptimization();
```
jeremyhyde-viam, line 489: [q] Doesn't running final optimization when extracting pointcloud introduce a time lag? I thought thats one reason why we have decided on using a optimization flag

slam-libraries/viam-cartographer/src/io/submap_painter.cc
```diff
@@ -113,5 +113,27 @@ cartographer::io::PaintSubmapSlicesResult PaintSubmapSlices(
                                                      origin);
 }
 
+// Draw a position on the current map
```
kkufieta, line 116: [nit] Use pose instead of position.
[opt] Suggestion: "Draw a marker for the current pose within the map"

slam-libraries/viam-cartographer/src/io/submap_painter.cc
kkufieta: You tested this, right? Was moving the origin of the map an issue?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -65,8 +65,9 @@ ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
         // paint map accordingly with or without the marker
         // included. Ticket: https://viam.atlassian.net/browse/DATA-657
         std::string jpeg_img = "";
+        bool marker_flag = request->include_robot_marker();
```
kkufieta, line 68: [q] Will we have to implement adding the marker to the pointcloud as well? Did we do that for orbslam? If so, should we create a ticket for this?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -477,6 +487,16 @@ void SLAMServiceImpl::RunSLAM() {
...
+        auto local_poses = map_builder.GetLocalSlamResultPoses();
+        if (local_poses.size() > 0) {
+            tmp_global_pose =
+                map_builder.GetGlobalPose(trajectory_id, local_poses.back());
+        }
+    }
+
+    {
+        std::lock_guard<std::mutex> lk(viam_response_mutex);
+        latest_global_pose = tmp_global_pose;
```
jeremyhyde-viam, line 499: [opt] setting latest_global_pose directly on line 192 otherwise latest_global_pose doesnt need to be updated

jeremyhyde-viam (approved):
Nit comments, otherwise looks good. Agree with Kat maybe not worth doing as long as it doesnt hurt our testing and development in the interim.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -185,6 +186,11 @@ std::string SLAMServiceImpl::PaintMap() {
         submap_poses =
             map_builder.map_builder_->pose_graph()->GetAllSubmapPoses();
     }
+    cartographer::transform::Rigid3d global_pose;
+    {
+        std::lock_guard<std::mutex> lk(viam_response_mutex);
+        global_pose = latest_global_pose;
+    }
```
kkufieta, line 193: [q] Can we move this into the `if (marker_flag)` statement below, such that we grab the `latest_global_pose` only when we want to add the pose marker?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: [opt] `pose_marker_flag` instead of `marker_flag`

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -65,8 +65,9 @@ ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
         // paint map accordingly with or without the marker
         // included. Ticket: https://viam.atlassian.net/browse/DATA-657
```
kkufieta, line 66: [nit] Remove this comment (lines 64-66)

slam-libraries/viam-cartographer/src/io/submap_painter.cc
kkufieta, line 117: [nit] Use the word "pose" instead of "position" to be consistent with the naming so far. `DrawPoseOnSurface` instead of `DrawPositionOnSurface`

slam-libraries/viam-cartographer/src/io/submap_painter.h
```diff
@@ -21,6 +21,10 @@ cartographer::io::PaintSubmapSlicesResult PaintSubmapSlices(
                    cartographer::io::SubmapSlice>& submaps,
     double resolution);
 
+void DrawPositionOnSurface(
```
kkufieta, line 24: [nit] `DrawPoseOnSurface`

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -174,7 +175,7 @@ void SLAMServiceImpl::SetUpMapBuilder() {
     map_builder.BuildMapBuilder();
 }
 
-std::string SLAMServiceImpl::PaintMap() {
+std::string SLAMServiceImpl::PaintMap(bool marker_flag) {
```
kkufieta, line 178: [opt] `pose_marker_flag`

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: This is for at the end of the dataset in offline mode. Not sure if we want to keep it here but was leaving it in anyways. Currently we would not be able to update our map while this optimization occurs, which has made testing fun. I think the trick is making GetMap() not rely on the `map_builder_mutex`, which we have an [investigation ticket](https://viam.atlassian.net/browse/DATA-668) for

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -94,7 +94,7 @@ class SLAMServiceImpl final : public SLAMService::Service {
     void OverwriteMapBuilderParameters();
 
     // PaintMap paints the map in jpeg format
-    std::string PaintMap();
+    std::string PaintMap(bool marker_flag);
```
kkufieta, line 97: [opt] `pose_marker_flag`

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: with how offline mode currently tries to optimize at the end of the run, we need to make sure our final pose gets updated after that optimization. With out this final update our mutex protected `latest_global_pose` ends up out of date.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: oh also this part isn't in the grpc GetMap() call, this is in `RunSLAM()`

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I think this is fine!

kkufieta (commented):
Nice work! I have mostly opts as well as a couple of nits and questions.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: we did not include the marker for the pointcloud, in our spec we only wrote it for the jpeg endpoint

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: sure, only thought is moving it out when we figure out how to have maps protected by this mutex as well

slam-libraries/viam-cartographer/src/io/submap_painter.cc
JohnN193: Only thing was keeping the name consistent with the GRPC position endpoint since these functions all relate to that(using the same protected variable) but I am fine with changing to pose

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
jeremyhyde-viam: We discussed when and where to do optimization, either at the beginning with a potential long delay before being able to actually start or at the end and requiring users to close out of slam and wait before closing out the server. The result of the conversation as I remember was to do it at the beginning in the optimize flag was true as we can't control a user's behavior at the end of a run. 

This code might be an artifact from our initial implementation of slam when we were always doing optimization at the end. If you dont want to handle it in this PR, I'd recommend making a ticket

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
jeremyhyde-viam: when a map exists wont local_poses.size always be greater than 0 so we always count on it being updated at the end of the run? And if you are worried about the mutex we can add it around the if statement. Unless I'm misunderstanding something

jeremyhyde-viam (commented):
More questions

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: yeah thats true, i can get rid of the check at the end here. We only need it at the start of a run/map

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: yeah I think it makes sense to move/change this on the ticket for implementing the optimize flag. Is the start/end behavior written down somewhere?

jeremyhyde-viam: Discussed here: https://docs.google.com/document/d/1LdD_q20timS7nUF43Aqa1fJ__p8N1u8C-NVVT6C0V9I/edit#bookmark=id.r6yew9ly8e33

slam-libraries/viam-cartographer/src/io/submap_painter.cc
JohnN193: did not appear to be an issue based off the images

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I don't see harm in saving an optimized map at the end of a successful offline run.

kkufieta: An example of this might be that I run through an offline dataset, and I want to see the optimized map - which looks different from before the optimization. Without optimizing the map at the end, there is no way for a user to view the optimized map.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: ah, I see now why you have it here. I think that's fine then.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
jeremyhyde-viam: I agree some what, however it goes against the tech spec so we need to discuss if we want to include it.

Options:
1. Do it and document that this occurs in offline but not online
2. Don't do it and require a user to load the map up and set optimize_on_start flag to perform operation and then save (not sure exactly how they would play out with no new data, although we did discuss doing it at the time)
3. introduce a optimize_at_end flag for explicit saving, offline or online

kkufieta: I think it should be option 1. I do also think that this should be put on another ticket rather than done in this PR because it isn't related to the task at hand.

kkufieta (approved):
LGTM!

PR #104, opened 2022-11-08
[DATA-713] Change map timestamp to new format

kkufieta:
Part of implementing this ticket: https://viam.atlassian.net/browse/DATA-713


JohnN193 (approved):
LGTM looks like integration tests are passing with this. Assuming you are testing with other changes locally

PR #106, opened 2022-11-10
[Data-708] Save pbstream maps at rate map_rate_sec

kkufieta:
https://viam.atlassian.net/browse/DATA-708


slam-libraries/viam-cartographer/src/io/file_handler.cc
```diff
@@ -19,6 +19,14 @@ namespace io {
...
 namespace fs = boost::filesystem;
 
+const std::string MakeFilenameWithTimestamp(std::string path_to_dir) {
+    std::time_t t = std::time(nullptr);
+    char timestamp[100];
+    std::strftime(timestamp, sizeof(timestamp), time_format.c_str(),
+                  std::gmtime(&t));
+    return path_to_dir + "/" + "map_data_" + timestamp + ".pbstream";
+}
+
```
kkufieta, line 29: Will test this function in this ticket: https://viam.atlassian.net/browse/DATA-681

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -332,11 +333,13 @@ bool SLAMServiceImpl::ExtractPointCloudToBuffer(std::stringbuf &buffer) {
     return has_points;
 }
 
-void SLAMServiceImpl::ProcessData() {
+void SLAMServiceImpl::RunSLAM() {
```
kkufieta, line 336: Swapped function names `RunSLAM` with `ProcessData` (now `ProcessDataAndStartSavingMaps`) cause they were more descriptive that way.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -380,8 +392,8 @@ std::string SLAMServiceImpl::GetNextDataFileOffline() {
     if (file_list_offline.size() == 0) {
         file_list_offline = viam::io::ListSortedFilesInDirectory(path_to_data);
     }
-    if (file_list_offline.size() == 0) {
-        throw std::runtime_error("no data in data directory");
+    if (file_list_offline.size() <= 2) {
+        throw std::runtime_error("not enough data in data directory");
```
kkufieta, line 396: Had to change this cause starting the mapping process with only one datafile produced a garbage map (and sometimes there's a `.DS_Store` file located in the directory so I set the minimum to 3 files)

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -420,7 +432,74 @@ std::string SLAMServiceImpl::GetNextDataFile() {
...
+    while (b_continue_session) {
+        auto start = std::chrono::high_resolution_clock::now();
+        const std::string filename_with_timestamp =
+            viam::io::MakeFilenameWithTimestamp(path_to_map);
+
+        if (offlineFlag && finished_processing_offline) {
+            {
+                std::lock_guard<std::mutex> lk(map_builder_mutex);
+                // Save the map in a pbstream file
+                bool ok = map_builder.map_builder_->SerializeStateToFile(
```
JohnN193, line 462: [opt] since this is duplicate code should we try to make it a function? just a void function to save the map should be fine, and leave the !ok check inside of it.

JohnN193 (approved):
LGTM

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -420,7 +432,74 @@ std::string SLAMServiceImpl::GetNextDataFile() {
...
+        // shutdown
+        while (b_continue_session) {
+            std::chrono::duration<double, std::milli> time_elapsed_msec =
+                std::chrono::high_resolution_clock::now() - start;
+            if ((time_elapsed_msec >= map_rate_sec) ||
+                (finished_processing_offline)) {
+                break;
+            }
+            if (map_rate_sec - time_elapsed_msec >=
+                check_for_shutdown_interval_usec) {
```
jeremyhyde-viam, line 491: @JohnN193 on shutdown timing

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -474,28 +567,41 @@ void SLAMServiceImpl::RunSLAM() {
...
         map_builder.map_builder_->FinishTrajectory(trajectory_id);
-        map_builder.map_builder_->pose_graph()->RunFinalOptimization();
-        auto local_poses = map_builder.GetLocalSlamResultPoses();
-        tmp_global_pose =
-            map_builder.GetGlobalPose(trajectory_id, local_poses.back());
     }
+    if (offlineFlag) {
+        {
+            std::lock_guard<std::mutex> lk(map_builder_mutex);
+            map_builder.map_builder_->pose_graph()->RunFinalOptimization();
```
jeremyhyde-viam, line 581: [minor] add log for optimizing start and end so people know whats happening

jeremyhyde-viam (commented):
Great addition Kat! A few minor notes but logic looks sound

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: Yes:
1) We should not be deleting files from people's directories. I suspect Macs create `.DS_Store` files for a purpose and we should not mess with that.
2) We should anticipate and handle automatically created files like that within our codebase robustly, which means making sure that we only work with `.pbstream` files directly and ignore everything else is the best course of action. In this piece of code I have to admit though that I'm making an assumption about how likely it is for us to encounter automatically added files - which I'm assuming for now only happens with `.DS_Store` files cause that was the first & only time we ran into this issue. I could write more robust code here that'll actually count the `.pbstream` files but I figured my assumption to be safe enough and it allows us to skip having to count the `.pbstream` files in the directory. If we start seeing more issues like these we can change this code to be more robust, as well.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: what is this about?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -436,20 +515,33 @@ void SLAMServiceImpl::RunSLAM() {
     bool set_start_time = false;
     auto file = GetNextDataFile();
 
-    // define tmp_global_pose here so it always has the previous pose
+    // Define tmp_global_pose here so it always has the previous pose
```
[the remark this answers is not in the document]
kkufieta: I think right now we seem to have a pattern of adding a . to `.h` file function descriptions (as we do when writing go code), but we seem to have no periods in regular comments. Will leave as is for consistency here

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -436,20 +515,33 @@ void SLAMServiceImpl::RunSLAM() {
...
     cartographer::transform::Rigid3d tmp_global_pose =
         cartographer::transform::Rigid3d();
     while (file != "") {
+        // Ignore files that are not *.pcd files
         if (file.find(".pcd") == std::string::npos) {
             file = GetNextDataFile();
             continue;
         }
+        // Go past files that are not supposed to be included in this run
+        if (!set_start_time &&
```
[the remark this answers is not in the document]
kkufieta: Agree this would be easier to read, but leaving it in the if statement so the `ReadTimeFromFilename` function will be only executed if `set_start_time != true`

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -49,15 +49,17 @@ class SLAMServiceImpl final : public SLAMService::Service {
...
 
-    // ProcessData process the data in the data directory. In offline mode,
+    // ProcessDataAndStartSavingMaps processes the data in the data directory
+    // that is newer than the provided data_cutoff_time
+    // and starts the process to save maps in parallel. In offline mode,
     // all data in the directory is processed. In online mode, the most
     // recently generated data is processed until a shutdown signal is
     // received.
-    void ProcessData();
+    void ProcessDataAndStartSavingMaps(double data_cutoff_time);
```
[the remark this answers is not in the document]
kkufieta: It is already split into those two functions internally. What this means is that `ProcessDataAndStartSavingMaps` handles both online & offline, and if you keep following the code you'll see eventually how online & offline has each a separate function.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -155,6 +169,9 @@ class SLAMServiceImpl final : public SLAMService::Service {
     size_t current_file_offline = 0;
     std::string current_file_online;
 
+    std::atomic<bool> finished_processing_offline{false};
+    std::thread *thread_save_map_with_timestamp;
```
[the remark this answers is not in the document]
kkufieta: This thread is for the `SaveMapWithTimestamp` function and it is easier to see what the thread is for if the thread's name contains the full corresponding function name.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: timing makes sense here. In orbslam Kat had setup the thread to loop at a faster rate in order to catch when a shutdown signal occurs, and this mirrors that implementation.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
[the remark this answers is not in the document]
jeremyhyde-viam: Then consider combining this `!set_start_time check` with the one on line 531 and having an internal if statement for `ReadTimeFromFilename`. 

Will result in the same number of if statements but will keep them simpler.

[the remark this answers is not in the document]
jeremyhyde-viam: also will provide a clean read for seeing the updating of the `file` variable

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
jeremyhyde-viam: Can we get a comment explaining the 2 then? I think its important to note why for future users

jeremyhyde-viam (approved):
Suggestion on addition comment and clean up of if statement. Will approve now as these are minor quality-of-life fixes

PR #107, opened 2022-11-10
[DATA-745] [DATA-761] build viam cartographer with local files and initial M1 setup

JohnN193:
Removes the automatic installation of cartographer & viam cartographer and builds using libraries on the local build path. In addition sets up a make target to grab dependencies for cartographer on macos and a make target to rebuild cartographer(to prevent building from scratch again)

https://viam.atlassian.net/browse/DATA-761

https://viam.atlassian.net/browse/DATA-745


slam-libraries/Makefile
```diff
@@ -60,13 +64,23 @@ appimage: buildorb
...
 	cd viam-cartographer/scripts && ./setup_cartographer.sh
+endif
+	
 
 buildcarto:
-	cd viam-cartographer/scripts && ./build_cartographer.sh && ./build_viam_cartographer.sh
+ifneq ($(wildcard viam-cartographer/build/.),)
+	cd viam-cartographer && ./scripts/rebuild_viam_cartographer.sh 
+else 
+	cd viam-cartographer && ./scripts/build_cartographer.sh && ./scripts/build_viam_cartographer.sh
```
Otterverse, line 78: The difference between "build" and "rebuild" scripts here looks to be just two lines. Not a good pattern to just copy/paste that much other stuff when you could have the IF statement just be in the script itself.

Otterverse (approved):
As long as this works on a clean environment, then that's good enough for me. I did make one note about some code reuse, but that's optional for now in my mind.

slam-libraries/Makefile
JohnN193: that makes sense, I can move the IF into the script

slam-libraries/viam-cartographer/scripts/rebuild_viam_cartographer.sh
```diff
@@ -13,6 +13,6 @@ fi
 
 # Rebuild viam cartographer.
 pushd build
+cmake .. -G Ninja -DCMAKE_CXX_STANDARD=17 -DCMAKE_PREFIX_PATH=`brew --prefix` -DQt5_DIR=$(brew --prefix qt5)/lib/cmake/Qt5
```
kkufieta, line 16: [q] Why did you add this line here? The idea of `rebuild_viam_cartographer.sh` is not having to run `cmake` again

kkufieta (changes requested):
Now that we have `setup_cartographer_macos.sh`, can you rename `setup_cartographer.sh` to something like `setup_cartographer_linux.sh` and update the README's to point to the new setup files? I believe the README for osx could now be updated to use `setup_cartographer_macos.sh` instead of the instructions to install libraries manually.

slam-libraries/viam-cartographer/scripts/rebuild_viam_cartographer.sh
JohnN193: if new files were added then the rebuild script would miss them. running the cmake script does not take very long. The reason why `build_viam_cartographer` takes longer is because we delete the build directory and start from scratch every time

slam-libraries/viam-cartographer/README.md
```diff
@@ -36,21 +46,21 @@ brew upgrade
 # Install dependencies
 brew install abseil boost ceres-solver protobuf ninja cairo googletest lua@5.3
 brew link lua@5.3
+brew link protobuf
+brew install openssl eigen gflags glog suite-sparse sphinx-doc pcl
```
kkufieta, line 50: Sorry - what I meant was the opposite: Rather than listing the dependencies manually, the user could just run `./scripts/setup_cartographer_macos.sh` here. I'd therefore replace this with:

```bash
# Install dependencies
./scripts/setup_cartographer_macos.sh
```

slam-libraries/viam-cartographer/README.md
```diff
@@ -36,21 +46,21 @@ brew upgrade
...
 ### Setup: Raspberry Pi OS Lite (64-bit)
 
-Run: `./scripts/setup_cartographer.sh`
- 
+**Install dependencies**
+```bash
+sudo apt update
+# Install dependencies
+sudo apt install -y cmake ninja-build libgmock-dev libboost-iostreams-dev liblua5.3-dev libcairo2-dev python3-sphinx libabsl-dev libceres-dev libprotobuf-dev protobuf-compiler libpcl-dev
+ ```
```
kkufieta, line 62: Replace this with:

```bash
# Install dependencies
./scripts/setup_cartographer_linux.sh
```

slam-libraries/viam-cartographer/README.md
```diff
@@ -72,13 +82,25 @@ Installation & building tested on:
...
-Configure how to run cartographer in this file: [scripts/run_cartographer.sh](./scripts/run_cartographer.sh).
 
-Run cartographer by executing: `./scripts/run_cartographer.sh`
+In your desired data directory, move the configuration files from cartographer into your `~/config` folder:  
+```bash
+cd viam-cartographer
+cp -r lua_files ~/YOUR_DATA_DIR/config/
+cp cartographer/configuration_files/* ~/YOUR_DATA_DIR/config/lua_files/
+```
+You only have to do this once per data directory. Note cartographer will fail if the configuration files cannot be found
```
kkufieta, line 100: Period after "found"

slam-libraries/viam-cartographer/README.md
```diff
@@ -72,13 +82,25 @@ Installation & building tested on:
...
+
+To run tests, after building, run the following
+```bash
+cd viam-cartographer
```
kkufieta, line 104: Remove `cd viam-cartographer` since we're already in that directory

slam-libraries/viam-cartographer/README.md
JohnN193: i added a block above for automatic installation. So either a user can use the auto install script, or get dependencies manually if they have specific versions of these packages that they want to use

slam-libraries/viam-cartographer/README.md
```diff
@@ -19,6 +19,16 @@ Learn more at [the official Docs site](https://google-cartographer.readthedocs.i
 ## Installation instructions
 Make sure to follow all steps as outlined in [the setup section here](../../README.md#setup).
 
+### Automatic Dependency Installation (x64, arm64, or macOS)
```
JohnN193, line 22: this section tells users they can automatically install dependencies if they want

slam-libraries/viam-cartographer/README.md
JohnN193: see above comment

slam-libraries/viam-cartographer/README.md
kkufieta: Right! ok makes sense

slam-libraries/viam-cartographer/README.md
JohnN193: tweaked wording abit to make it clear this is for manual installation, and we want to keep these because it shows users what is actually getting installed on their systems

PR #108, opened 2022-11-11
[DATA-733] Throw error if no map is found

kkufieta:
https://viam.atlassian.net/browse/DATA-733

**Situation:** The user chooses `map_rate_sec = 0` - which technically means running SLAM in localization action mode - but no map is found in the `data/map` folder.

**Previous behavior:** Run SLAM in mapping mode (since no map is found) but don't save maps (since `map_rate_sec == 0`).

**New behavior:** Throw an error, since we assume the user wants to run SLAM in localization mode but we can't find a map to localize on.

**Tested on:**
* 2.4 GHz 8-Core Intel Core i9; macOS Monterey

### DIFF
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -123,6 +123,10 @@ void SLAMServiceImpl::DetermineActionMode() {
             return;
         }
     }
+    if (map_rate_sec.count() == 0) {
+        throw std::runtime_error(
+            "In localization mode but couldn't find a map to localize on");
+    }
     LOG(INFO) << "Running in mapping mode";
     action_mode = SLAMServiceActionMode::MAPPING;
 }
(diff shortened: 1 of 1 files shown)


JohnN193 (approved):

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -123,6 +123,10 @@ void SLAMServiceImpl::DetermineActionMode() {
             return;
         }
     }
+    if (map_rate_sec.count() == 0) {
+        throw std::runtime_error(
+            "In localization mode but couldn't find a map to localize on");
```
jeremyhyde-viam, line 128: [opt] 
`map_rate_sec set to localization mode but couldn't find an apriori map`
or
`set to localization mode (map_rate_sec = 0) but couldn't find apriori map to localize on`

jeremyhyde-viam (approved):
LGTM, just suggested a change to the error thrown to be a bit more explicit about how localization mode was determined

PR #111, opened 2022-11-18
[DATA-124]  [DATA-787] create cartographer appimage

JohnN193:
Adds appimages for cartographer and sets up CI to use them. Additionally has cartographer check for lua files added by the appimage.

Successfully built in CI(with changes from PR review): https://github.com/viamrobotics/slam/actions/runs/3512196057

tested on a clean PI & an x86 linux machine

https://viam.atlassian.net/browse/DATA-124 - create appimage
https://viam.atlassian.net/browse/DATA-787 - find lua file from appimage


slam-libraries/Makefile
```diff
@@ -56,13 +56,20 @@ testorb:
...
 
-appimage: buildorb
+appimageorb: buildorb
 	cd packaging/appimages && BUILD_CHANNEL=${BUILD_CHANNEL} appimage-builder --recipe orb_grpc_server-`uname -m`.yml
 	cd packaging/appimages && ./package_release.sh
 	mkdir -p packaging/appimages/deploy/
 	mv packaging/appimages/*.AppImage* packaging/appimages/deploy/
 	chmod 755 packaging/appimages/deploy/*.AppImage
 
+appimagecarto: buildcarto
```
JohnN193, line 66: separated the appimages so they can build separately in CI

slam-libraries/viam-cartographer/src/slam_service/config.cc
```diff
@@ -67,6 +70,24 @@ void ParseAndValidateConfigParams(int argc, char** argv,
     slamService.path_to_data = FLAGS_data_dir + "/data";
     slamService.path_to_map = FLAGS_data_dir + "/map";
     slamService.configuration_directory = FLAGS_data_dir + "/config/lua_files";
+
+    // Find the lua files.
+    path pathToLuasFromConfig(slamService.configuration_directory);
```
JohnN193, line 75: This code looks for lua files being placed in our appimage recipe, so users that have an appimage do not need to place the files into their config directory

.github/workflows/appimage.yml
```diff
@@ -103,7 +103,106 @@ jobs:
         sudo --preserve-env=GOOGLE_APPLICATION_CREDENTIALS -u testbot bash -lc 'cd rdk/services/slam/builtin && sudo go test -v -run TestOrbslamIntegration'
 
     - name: Build appimage
-      run: sudo -u testbot bash -lc 'cd slam/slam-libraries && make BUILD_CHANNEL="latest" appimage'
+      run: sudo -u testbot bash -lc 'cd slam/slam-libraries && make BUILD_CHANNEL="latest" appimageorb'
+
+    - name: Authorize GCP Upload
+      uses: google-github-actions/auth@v0.4.3
```
Otterverse, line 109: Switch to @v1 here rather than pinning an out-of-date exact version. Actually, maybe just double-check all the workflows while your at it for that. (I know this is used in other jobs in the workflow too.)

.github/workflows/appimage.yml
```diff
@@ -103,7 +103,106 @@ jobs:
...
+      with:
+        credentials_json: '${{ secrets.GCP_CREDENTIALS }}'
+    
+    - name: Deploy Files
+      uses: google-github-actions/upload-cloud-storage@v0.5.0
```
Otterverse, line 114: @v1 here also.

Otterverse (approved):
Looks good, save for the version bumps I noted. Gonna go ahead and approve though, and trust you to take of that before merging. (Do test it first, just in case the new versions give you any trouble.)

.github/workflows/appimage.yml
JohnN193: Got everything in the slam repo but looks like camera-servers is out of date as well, I'll go ahead and make a PR for that

Otterverse: I don't think it's critical enough to need a PR right away or anything. It was just more of a "while you're in here..."

PR #112, opened 2022-11-22
[DATA-681] Add tests for file_handler.cpp

kkufieta:
Write tests for file_handler.cpp

https://viam.atlassian.net/browse/DATA-681


slam-libraries/viam-cartographer/src/io/file_handler_test.cc
```diff
@@ -0,0 +1,183 @@
...
+
+BOOST_AUTO_TEST_CASE(MakeFilenameWithTimestamp_success) {
+    std::string path_to_dir = "path_to_dir";
+    std::time_t start_time =
+        std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+    std::string filename = MakeFilenameWithTimestamp(path_to_dir);
+    std::time_t end_time =
+        std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+    // Check if the filename beginning is as expected
+    std::string path_appendix = "/map_data_";
```
jeremyhyde-viam, line 23: [nit] appendix is usually at the end of the associated object, use prefix if you are refering to the front

slam-libraries/viam-cartographer/src/io/file_handler_test.cc
```diff
@@ -0,0 +1,183 @@
...
+    double filename_time = ReadTimeFromFilename(filename.substr(
+        filename.find(filename_prefix) + filename_prefix.length(),
+        filename.find(".pcd")));
+    // Check if timestamp is between start_time and end_time
+    BOOST_TEST((double)start_time <= filename_time);
+    BOOST_TEST(filename_time >= (double)end_time);
+}
+
+BOOST_AUTO_TEST_CASE(ListSortedFilesInDirectory_success) {
+    // Create a temp directory with a few files with timestamps
```
jeremyhyde-viam, line 37: [nit] a few 'sorted' files with timestamps

slam-libraries/viam-cartographer/src/io/file_handler_test.cc
```diff
@@ -0,0 +1,183 @@
...
+    if (!ok) {
+        throw std::runtime_error("could not create directory: " +
+                                 tmpdir.string());
+    }
+    for (std::string file : files) {
+        boost::filesystem::ofstream ofs(tmpdir / file);
+        ofs.close();
+    }
+    // Remove a file
+    int success = RemoveFile(tmpdir.string() + "/" + files.at(1));
```
jeremyhyde-viam, line 89: [opt] could replace the one with a file_num variable that could be reused on line 91

slam-libraries/viam-cartographer/src/io/file_handler_test.cc
```diff
@@ -0,0 +1,183 @@
...
+    BOOST_TEST(timed_pcd.ranges.size() == 3);
+    cartographer::sensor::TimedRangefinderPoint point_1 =
+        timed_pcd.ranges.at(0);
+    cartographer::sensor::TimedRangefinderPoint point_2 =
+        timed_pcd.ranges.at(1);
+    cartographer::sensor::TimedRangefinderPoint point_3 =
+        timed_pcd.ranges.at(2);
+
+    auto tolerance = boost::test_tools::tolerance(0.00001);
+    BOOST_TEST(point_1.position(0, 0) == -0.001, tolerance);
```
jeremyhyde-viam, line 146: [opt] similar to above comment, I would make a list of the three pc values that then would be placed in the pcd file and could be referenced below so that if the test changes in the future it may only need to be edited in one place.

slam-libraries/viam-cartographer/src/io/file_handler_test.cc
```diff
@@ -0,0 +1,183 @@
...
+    BOOST_TEST(point_2.position(2, 0) == 0.000, tolerance);
+
+    BOOST_TEST(point_3.position(0, 0) == 0.007, tolerance);
+    BOOST_TEST(point_3.position(1, 0) == 0.006, tolerance);
+    BOOST_TEST(point_3.position(2, 0) == 0.001, tolerance);
+    // Close the file and remove the temporary directory and its contents
+    boost::filesystem::remove_all(tmpdir);
+}
+
+BOOST_AUTO_TEST_CASE(ReadTimeFromFilename_success) {
```
jeremyhyde-viam, line 161: [q] Do we want to test a failed case of this? Or are there too many ways a data filename might be wrong to make it worth testing?

jeremyhyde-viam (approved):
LGTM, a few minor improvements that would make adding or changing tests a more seamless process

slam-libraries/viam-cartographer/src/io/file_handler_test.cc
JohnN193: Maybe a check with a filename that has an incorrect format?

JohnN193 (approved):
LGTM

slam-libraries/viam-cartographer/src/io/file_handler_test.cc
kkufieta: good idea, adding

PR #113, opened 2022-11-29
[DATA-117] Read optimize_on_start parameter from config_params

kkufieta:
Ticket: https://viam.atlassian.net/browse/DATA-117

Done:
* Read optimize_on_start parameter from config parameters and use that to decide whether or not an uploaded map should be optimized before using it
* Rename: `offlineFlag` --> `offline_flag` for consistency

### DIFF
--- a/slam-libraries/viam-cartographer/src/slam_service/config.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/config.cc
@@ -45,12 +45,6 @@ void ParseAndValidateConfigParams(int argc, char** argv,
     if (!v.empty()) {
         FLAGS_v = std::stoi(v);
     }
-    const auto optimize_on_start =
-        ConfigParamParser(FLAGS_config_param, "optimize_on_start=");
-    if (optimize_on_start == "true") {
-        slamService.optimize_on_start = true;
-        LOG(INFO) << "Optimizing map on start";
-    }
 
     if (FLAGS_data_dir.empty()) {
         throw std::runtime_error("-data_dir is missing");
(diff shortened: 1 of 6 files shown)


jeremyhyde-viam (approved):
LGTM, a few minor suggestions and check on what the reset value for optimize_on_start should be.

slam-libraries/viam-cartographer/src/slam_service/config_test.cc
```diff
@@ -226,6 +227,7 @@ BOOST_AUTO_TEST_CASE(
     ResetFlagsForTesting();
     std::string config_param =
         "{mode=2d,"
+        "optimize_on_start=true,"
```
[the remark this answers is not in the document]
kkufieta: I don't understand what you're asking about. I'm setting `optimize_on_start` purposefully to `true` here. Resetting flags happens in `ResetFlagsForTesting()` above that. Can you explain in more detail what you're curious about?

slam-libraries/viam-cartographer/src/slam_service/config.cc
```diff
@@ -84,6 +84,13 @@ void ParseAndValidateConfigParams(int argc, char** argv,
         throw std::runtime_error("Invalid slam_mode=" + slamService.slam_mode);
     }
 
+    const auto optimize_on_start =
+        ConfigParamParser(FLAGS_config_param, "optimize_on_start=");
+    if (optimize_on_start == "true") {
+        slamService.optimize_on_start = true;
+        LOG(INFO) << "Optimizing map on start, this may take a few minutes";
```
JohnN193, line 91: [q] should this log still appear if there is no map to optimize?

JohnN193 (approved):
LGTM, just one question

slam-libraries/viam-cartographer/src/slam_service/config.cc
kkufieta: Very good point. Moved this line into `map_builder.cc` so that it's only called before the optimization actually starts.

PR #114, opened 2022-11-29
[DATA-736] Cartographer: Update warnings at top of files & remove main_old

kkufieta:
Ticket: https://viam.atlassian.net/browse/DATA-736

Done:
* Updated warnings at top of the files to match the warnings of orbslam
* Removed outdated warning from readme
* Removed main_old.cpp and the related script


slam-libraries/viam-cartographer/README.md
```diff
@@ -5,9 +5,6 @@
 > **Warning**
 > This is an experimental feature. Stability is not guaranteed. Breaking changes are likely to occur, and occur often.
 
-> **Disclaimer**
-> Cartographer has not yet been integrated into RDK.
```
jeremyhyde-viam, line 9: [minor] `Cartographer has not yet been officially integrated into RDK.`

slam-libraries/viam-cartographer/src/io/color.cc
```diff
@@ -1,5 +1,4 @@
-// This is an Experimental variation of cartographer. It has not yet been
-// integrated into RDK.
+// This is an Experimental integration of cartographer into RDK.
```
jeremyhyde-viam, line 1: [q] why is experimental capitalized?

jeremyhyde-viam (commented):
Two minor questions to answer before sign off

slam-libraries/viam-cartographer/src/io/color.cc
kkufieta: Cause that's how it's done in orbslam and while I had the same question as you, I just copy/pasted it. Don't like it, though, and the fact that you pointed it out changed my mind about it. I'll go and fix it for both cartographer and orbslam.

JohnN193: I’m guessing it’s because Experimental is the name of the label that should be defined somewhere in Viam documentation and not purely an adjective

kkufieta: It's lowercase in the documentation:
* https://docs.viam.com/services/slam/
* https://docs.viam.com/tutorials/slam/

Which is why I think it was probably an artifact that originated from somewhere else (maybe a typo in the ticket description).

slam-libraries/viam-cartographer/README.md
kkufieta: There are two reasons why I can't entirely agree with making this change:
1. Cartographer is integrated into RDK since the user can follow the instructions to install cartographer and run it through RDK.
2. We're currently wrapping up the last few tickets to finish cartographer, and this task is one of them. If we were to make the change you suggested, we'd have to create another ticket for undoing it later again. I don't think that's worth the effort.

jeremyhyde-viam (approved):
Approved!

PR #116, opened 2022-12-05
[DATA-661] Stop attempting to paint new maps when optimizing the map

kkufieta:
Ticket: https://viam.atlassian.net/browse/DATA-661

**Problem:**
We optimize the map in two situations:
* At startup, when an apriori map is loaded and `optimize_on_start = true`.
* At the end of processing an offline dataset.

The `map_builder_mutex` is held while the optimization is happening, so grabbing the map in `jpeg` or `pcd` form would have to wait until the optimization is finished. This results in the user having to wait a long time for the map to arrive after requesting it. 

**Solution:**
This solution checks whether or not the optimization is ongoing, and if so, `GetMap` will return the most recently saved map. 

Reasoning: The map is not changing while the optimization is ongoing. As a result, if the user requests a map during the optimization, we can return the most recently created map. 

**Comparing times:**
I tested and timed mapping before and after the changes I've made to ensure that mapping and requesting & receiving maps do not slow down. I used the [viam-old-office-small-pcd](https://storage.googleapis.com/viam-labs-datasets/viam-old-office-small-pcd.zip) dataset.

The times for "before" --> "after" the changes I made are reduced on average:
* Total runtime for mapping the entire dataset: 25 seconds --> 25 seconds
* `slam::server::GetMap`: 784 ms --> 501 ms
* `slam::builtin::GetMap`: 698 ms --> 416 ms


slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -359,10 +361,12 @@ void SLAMServiceImpl::RunSLAM() {
         bool load_frozen_trajectory =
             action_mode == SLAMServiceActionMode::LOCALIZING;
         {
+            optimizing = (optimize_on_start == true);
             // Load apriori map
             std::lock_guard<std::mutex> lk(map_builder_mutex);
             map_builder.LoadMapFromFile(
                 latest_map_filename, load_frozen_trajectory, optimize_on_start);
+            optimizing = false;
```
tessavitabile, line 369: We need to make sure `optimizing` gets set to `false` even if the above code throws, so the above code should be wrapped in a try/catch.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -578,6 +583,7 @@ void SLAMServiceImpl::ProcessDataAndStartSavingMaps(double data_start_time) {
                 tmp_global_pose = map_builder.GetGlobalPose(trajectory_id,
                                                             local_poses.back());
             }
+            optimizing = false;
```
tessavitabile, line 586: Same here

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -171,6 +171,11 @@ class SLAMServiceImpl final : public SLAMService::Service {
     size_t current_file_offline = 0;
     std::string current_file_online;
 
+    std::atomic<bool> optimizing{false};
+    std::string jpeg_img = "";
```
tessavitabile, line 175: `jpeg_img` and `buffer` need to be mutex-protected, so it is safe to access them concurrently.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -171,6 +171,11 @@ class SLAMServiceImpl final : public SLAMService::Service {
...
+    std::atomic<bool> has_points{false};
+    std::stringbuf buffer;
```
tessavitabile, line 177: I would choose a more descriptive name, such as `pointcloud_buffer`.

tessavitabile (changes requested):
Nice work investigating this! One concern I have is that the staleness of the map depends on the last time `GetMap` was called. This could be fixed by updating `jpeg_img` and `buffer` each time the map is updated. If we take that approach, then it would be beneficial to always have `GetMap` return the most recent map, not just in the case that the map is currently optimizing. This would also allow us to stop taking `map_builder_mutex` in `GetMap`. What do you think?

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
kkufieta: I thought they didn't need to be protected since they're only read and written within `GetMap`. My assumption was that subsequent calls to `GetMap` would be executed sequentially rather than concurrently. Was I mistaken in my assumption?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I don't understand. The way I think about this is: If the above code throws, I would assume that cartographer will error out and will not be executed, which is what we want. In that case, we don't care what happens after `map_builder.LoadMapFromFile(...)`.

Am I getting this wrong? How do you think about this?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I can see how using a try/catch here might make sense. This way we could continue serving requests even if the optimization fails. Is this the motivation for using a try/catch here?

If so, then my concern is that if the optimization fails, the map might be corrupted, and we'd prefer to have cartographer fail and stop serving requests after all. What do you think about that?

kkufieta (commented):
@tessavitabile 

> One concern I have is that the staleness of the map depends on the last time GetMap was called. 

I agree with this concern, especially when the time between requests is large. We might be able to update the most recent copy of the map before starting the optimization.

> This could be fixed by updating jpeg_img and buffer each time the map is updated. If we take that approach, then it would be beneficial to always have GetMap return the most recent map, not just in the case that the map is currently optimizing. This would also allow us to stop taking map_builder_mutex in GetMap. What do you think?

I am almost certain this will slow down cartographer considerably since painting the map to obtain a jpeg version takes a long time. We should update the maps as few times as possible, which in this case is every time the user requests it.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: The motivation for the try/catch is to make sure we still set `optimizing` to the correct value. That's true that if the exception terminates the program, then it's not important. But if the calling code catches the exception, `optimizing` should be set correctly. I think the correct way to do this is wrap in a try/catch and re-throw the original exception in the catch.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
tessavitabile: If there are multiple app instances view the same robot, there could be concurrent calls to `GetMap`.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -63,10 +63,11 @@ ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
     response->set_mime_type(mime_type);
 
     if (mime_type == "image/jpeg") {
-        std::string jpeg_img = "";
         bool pose_marker_flag = request->include_robot_marker();
         try {
-            jpeg_img = PaintMap(pose_marker_flag);
+            if (!optimizing) {
```
tessavitabile, line 68: The value of `optimizing` could change immediately after this check, so I think `optimizing` needs to be a mutex instead. Then this code would be 

try to lock mutex
if fail, use existing jpeg_image
if succeed, keep mutex locked during call to PaintMap

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -86,8 +87,9 @@ ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
             return grpc::Status(grpc::StatusCode::UNAVAILABLE, oss.str());
         }
     } else if (mime_type == "pointcloud/pcd") {
-        std::stringbuf buffer;
-        bool has_points = ExtractPointCloudToBuffer(buffer);
+        if (!optimizing) {
```
tessavitabile, line 90: Same here

tessavitabile (changes requested):
> I agree with this concern, especially when the time between requests is large. We might be able to update the most recent copy of the map before starting the optimization.

Updating the most recent copy of the map before starting the optimization makes sense to me. In that case, we don't need to update the map on calls to `GetMap`.

> I am almost certain this will slow down cartographer considerably since painting the map to obtain a jpeg version takes a long time. We should update the maps as few times as possible, which in this case is every time the user requests it.

Okay, sounds good.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: oooh that's smart! I like that.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -171,6 +171,11 @@ class SLAMServiceImpl final : public SLAMService::Service {
     size_t current_file_offline = 0;
     std::string current_file_online;
 
+    std::atomic<bool> optimizing{false};
+    std::string jpeg_img = "";
+    std::atomic<bool> has_points{false};
+    std::stringbuf pointcloud_buffer;
```
jeremyhyde-viam, line 177: [q]q whats the benefit of saving pointcloud_buffer as a global variable instead of passing it through as before? I tend to like to use globals as minimally as possible to better understand the scope of that var.

jeremyhyde-viam (commented):
A few questions before sign off, but found no major problems

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -568,6 +572,7 @@ void SLAMServiceImpl::ProcessDataAndStartSavingMaps(double data_start_time) {
     }
     if (offline_flag) {
         {
+            optimizing = true;
```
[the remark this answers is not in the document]
kkufieta: yes, that's correct

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: Using a mutex instead that is released when the lock goes out of scope.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: Using a mutex instead that is released when the lock goes out of scope.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
kkufieta: We need to save it as a member variable since we need to save the map as a backup that `GetMap` can use to return a map while cartographer's optimization process blocks the `map_builder_mutex`.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -162,24 +145,88 @@ class SLAMServiceImpl final : public SLAMService::Service {
...
         cartographer::transform::Rigid3d();
+
+    // --- The following variables are used exclusively to
+    // enable GetMap to send the most recent map out while
+    // cartographer works on creating an optimized map.
+    // The variables are only updated right before the
+    // optimization is started.
+    std::shared_mutex optimization_shared_mutex;
+
+    std::mutex latest_jpeg_map_str_mutex;
```
tessavitabile, line 222: [opt] I prefer combining `latest_global_pose_mutex`, `latest_jpeg_map_str_mutex`, and `latest_pointcloud_map_mutex` into a single `viam_response_mutex`, since these are used to control member variable accesses, which should all be fast. To me, that's less to think about, and it is more idiomatic to have a single mutex protecting member variable accesses.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -162,24 +145,88 @@ class SLAMServiceImpl final : public SLAMService::Service {
...
+    // received.
+    void ProcessDataAndStartSavingMaps(double data_cutoff_time);
+
+    // SetUpMapBuilder loads the lua file with default cartographer config
+    // parameters depending on the action mode.
+    void SetUpMapBuilder();
+
+    // GetLatestJpegMapString paints and returns the latest map as a jpeg string
+    // with or without the pose marker depending on the argument value.
+    std::string GetLatestJpegMapString(bool add_pose_marker);
```
tessavitabile, line 176: Please add comments to functions that are expected to be called with a particular mutex held.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -179,8 +230,18 @@ void SLAMServiceImpl::SetUpMapBuilder() {
...
+    cartographer::io::PaintSubmapSlicesResult painted_slices =
+        viam::io::PaintSubmapSlices(submap_slices, kPixelSize);
+    if (add_pose_marker) {
+        PaintMarker(painted_slices);
+    }
+    return viam::io::WritePaintedSlicesToJpegString(painted_slices);
+}
+
+std::map<cartographer::mapping::SubmapId, ::cartographer::io::SubmapSlice>
+SLAMServiceImpl::GetLatestSubmapSlices() {
```
tessavitabile, line 244: [opt] I think it's clearer to keep this as part of `GetLatestJpegMapString`.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -63,78 +59,133 @@ ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
...
+                                           GetMapResponse *response) {
+    std::string jpeg_map_str = "";
+    bool add_pose_marker = request->include_robot_marker();
+
+    // Paint or grab the latest occupancy map in form of a jpeg string
+    try {
+        {
+            std::shared_lock optimization_lock{optimization_shared_mutex,
+                                               std::defer_lock};
+            if (optimization_lock.try_lock()) {
```
tessavitabile, line 83: This should be `try_lock_shared`

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -63,78 +59,133 @@ ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
...
-SLAMServiceActionMode SLAMServiceImpl::GetActionMode() { return action_mode; }
+::grpc::Status SLAMServiceImpl::GetPointCloudMap(const GetMapRequest *request,
+                                                 GetMapResponse *response) {
+    bool pointcloud_has_points = false;
+    std::string pointcloud_map_str;
+    // Write or grab the latest pointcloud map in form of a string
+    try {
+        std::shared_lock optimization_lock{optimization_shared_mutex,
+                                           std::defer_lock};
+        if (optimization_lock.try_lock()) {
```
tessavitabile, line 128: This should be `try_lock_shared`

tessavitabile (changes requested):
Great work on the concurrency! Have you repeated your test of calling GetMap while optimization is in progress?

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -162,24 +145,88 @@ class SLAMServiceImpl final : public SLAMService::Service {
...
     const std::string configuration_update_basename = "updating_a_map.lua";
+
     std::vector<std::string> file_list_offline;
     size_t current_file_offline = 0;
     std::string current_file_online;
 
-    std::atomic<bool> finished_processing_offline{false};
-    std::thread *thread_save_map_with_timestamp;
-
     std::mutex map_builder_mutex;
```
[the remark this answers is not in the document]
kkufieta: What do you mean? Can you give me an example?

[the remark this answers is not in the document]
tessavitabile: E.g. If mutexes A and B are held concurrently, then mutex A must be taken before mutex B. No other mutexes are expected to be held concurrently.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -23,24 +24,19 @@ ::grpc::Status SLAMServiceImpl::GetPosition(ServerContext *context,
...
     PoseInFrame *inFrame = response->mutable_pose();
     Pose *myPose = inFrame->mutable_pose();
 
-    // set pose for our response
+    // Set pose for our response
     myPose->set_x(global_pose.translation().x());
     myPose->set_y(global_pose.translation().y());
     myPose->set_z(global_pose.translation().z());
 
-    // TODO DATA-531: Remove extraction and conversion of quaternion from the
```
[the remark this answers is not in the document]
kkufieta: Because the [ticket](https://viam.atlassian.net/browse/DATA-531) contains the same information as this TODO comment, it describes very well that the change has to happen in this location. I don't think we should keep `TODO` comments around if they're not adding any value or if they're not serving as a placeholder for code that still has to be implemented. In this case, the comment is mostly a distraction that makes the code harder to read.

slam-libraries/viam-cartographer/src/io/submap_painter.cc
```diff
@@ -134,5 +135,11 @@ void DrawPoseOnSurface(
     cairo_surface_flush(painted_slices->surface.get());
 }
 
+std::string WritePaintedSlicesToJpegString(
+    cartographer::io::PaintSubmapSlicesResult& painted_slices) {
+    auto image = Image(std::move(painted_slices.surface));
```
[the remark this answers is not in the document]
kkufieta: Good call, I haven't thought about this. I'll rewrite this such that this is not going to be an issue.

kkufieta (commented):
> Have you repeated your test of calling GetMap while optimization is in progress?

I have! The total runtime is still unchanged at ~25 seconds.

I calculated the average time for the `GetMap` calls for two separate runs, and they're still below the time it took for the calls before I made these changes:

| | `slam::server::GetMap` | `slam::builtin::GetMap` |
| --- | --- | --- |
| Before these changes | 784 ms | 698 ms |
| After: 1st run | 380 ms | 296 ms |
| After: 2nd run | 517 ms | 436 ms|

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
tessavitabile: I thought it is required that `optimization_shared_mutex` is held while calling this function, since otherwise this function is unacceptably slow.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: Thanks, you are correct.

tessavitabile (approved):
LGTM! I'm not very concerned about the outcome of my open comment.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
kkufieta: Ah! The function is slow but can be called irrespectively `optimization_shared_mutex`. The `optimization_shared_mutex` checks merely when the optimization is running - and if it is running, then we're returning the maps we generated and saved before the optimization started.

PR #117, opened 2022-12-09
DATA-723 Set Max Message Size for SLAM to 32 MB (SLAM)

jeremyhyde-viam:
gRPC limits the max size of bytes that can be sent to 4MB by default. The limit in most of RDK is 32MB however this also must be set in the SLAM algorithm's grpc system.

JIRA Ticket: [DATA-723](https://viam.atlassian.net/browse/DATA-723)
Associated PR: [#1658](https://github.com/viamrobotics/rdk/pull/1658)


JohnN193 (approved):

JohnN193 (approved):
LGTM, only question is does the maxMessageSize need to be a global variable?

kkufieta (approved):
LGTM - though I agree with John, have it inline and add a comment as to why we chose 32 MB, e.g., like this:

```cpp
// Increasing the gRPC max message size from the default value of 4MB to 32MB,
// to match the limit that is set in RDK. This is necessary for transmitting large
// pointclouds.
builder.SetMaxSendMessageSize(32 * 1024 * 1024);
```

PR #118, opened 2022-12-14
[DATA-693] update missing include

JohnN193:
When James tested the initial brew package for cartographer, he ran into an error as documented [here](https://github.com/viamrobotics/homebrew-brews/pull/8#pullrequestreview-1210828541).


kkufieta (approved):
LGTM

PR #120, opened 2022-12-19
RSDK-810 Delete Processed Data in SLAM (SLAM)

jeremyhyde-viam:
Adds the ability to remove files after processing. Current implementation according to scope doc (linked below) has offline mode always keep data, and online mode can be set to true to delete data files after they have been processed.

JIRA Ticket: [RSDK-1040](https://viam.atlassian.net/browse/RSDK-1040)
Associated PR: [#1689](https://github.com/viamrobotics/rdk/pull/1689)
Scope Doc: [Scope: SLAM usability improvements](https://docs.google.com/document/d/1RsT-c0QOtkMKa-rwUGY0-emUmKIRSaO3mzT1v6MtFjk/edit#)


slam-libraries/viam-cartographer/src/slam_service/config_test.cc
```diff
@@ -334,6 +334,56 @@ BOOST_AUTO_TEST_CASE(ParseAndValidateConfigParams_valid_config_no_camera) {
     delete argv;
 }
 
+BOOST_AUTO_TEST_CASE(
```
jeremyhyde-viam, line 337: can incorporate into other tests if needed, but felt it nice to break out.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -764,6 +778,14 @@ void ParseAndValidateArguments(const vector<string> &args,
         BOOST_LOG_TRIVIAL(info) << "No camera given -> running in offline mode";
         slamService.offlineFlag = true;
     }
+
+    auto delete_processed_data = ArgParser(args, "-delete_processed_data=");
+    slamService.delete_processed_data = !(delete_processed_data == "false");
```
jeremyhyde-viam, line 783: This will only change the value from the default of true if false is given. If anything else, say 'blahblahblah' is given, value will remain true. Could add an additional error throwing check if the value of `delete_processed_data != 'false' || delete_processed_data != 'true'`

slam-libraries/viam-cartographer/src/slam_service/config_test.cc
jeremyhyde-viam: See orbslam tests for example of including in current BOOST tests

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -119,5 +121,8 @@ int FindFrameIndex(const std::vector<std::string> &filesRGB,
 // currently does not support millisecond resolution
 string MakeFilenameWithTimestamp(string path_to_dir, string camera_name);
 
+// Removes data file
+int RemoveFile(std::string file_path);
```
jeremyhyde-viam, line 125: int is return to match RemoveFile previously defined in cartographer. Do we want to check and throw error if delete file fails or just ignore return?

jeremyhyde-viam (commented):
A few questions and comments on the existing implementation

slam-libraries/viam-cartographer/src/slam_service/config.cc
```diff
@@ -30,6 +30,12 @@ DEFINE_int64(
     "Frequency at which we want to print map pictures while cartographer "
     "is running.");
 DEFINE_string(input_file_pattern, "", "Input file pattern");
+// TODO: Switch back to bool once PR #1689 is submitted
```
tessavitabile, line 33: Is the plan to merge this PR, followed by #1689, followed by a PR to change this to a bool (which makes it required)?

slam-libraries/viam-cartographer/src/slam_service/config.cc
```diff
@@ -94,6 +102,21 @@ void ParseAndValidateConfigParams(int argc, char** argv,
     slamService.data_rate_ms = std::chrono::milliseconds(FLAGS_data_rate_ms);
     slamService.map_rate_sec = std::chrono::seconds(FLAGS_map_rate_sec);
 
+    // TODO: Remove once PR #1689 is submitted
+    // https://github.com/viamrobotics/rdk/pull/1689 This will allow intergation
+    if (FLAGS_delete_processed_data.empty()) {
+        slamService.delete_processed_data = false;
+        LOG(ERROR) << "HI\n";
```
tessavitabile, line 109: remove

slam-libraries/viam-cartographer/src/slam_service/config.cc
```diff
@@ -94,6 +102,21 @@ void ParseAndValidateConfigParams(int argc, char** argv,
...
+    } else {
+        LOG(ERROR) << "HO\n";
```
tessavitabile, line 111: remove

slam-libraries/viam-cartographer/src/slam_service/config.cc
```diff
@@ -94,6 +102,21 @@ void ParseAndValidateConfigParams(int argc, char** argv,
...
+        slamService.delete_processed_data =
+            !(FLAGS_delete_processed_data == "false");
+    }
+    // slamService.delete_processed_data = FLAGS_delete_processed_data;
+    // if (slamService.offline_flag) {
+    //     slamService.delete_processed_data = false;
```
tessavitabile, line 117: This can be an error, since rdk should never pass this combination of parameters.

slam-libraries/viam-cartographer/src/slam_service/config.cc
```diff
@@ -30,6 +30,12 @@ DEFINE_int64(
     "Frequency at which we want to print map pictures while cartographer "
     "is running.");
 DEFINE_string(input_file_pattern, "", "Input file pattern");
+// TODO: Switch back to bool once PR #1689 is submitted
+// https://github.com/viamrobotics/rdk/pull/1689 This will allow intergation
```
tessavitabile, line 34: I think this comment trailed off here, and in a couple other places.

slam-libraries/viam-cartographer/src/slam_service/config_test.cc
```diff
@@ -334,6 +334,58 @@ BOOST_AUTO_TEST_CASE(ParseAndValidateConfigParams_valid_config_no_camera) {
...
+    ParseAndValidateConfigParams(argc, argv, slamService);
+    BOOST_TEST(slamService.offline_flag == true);
+    BOOST_TEST(slamService.delete_processed_data == false);
+    delete argv;
+}
+
+BOOST_AUTO_TEST_CASE(
+    ParseAndValidateConfigParams_valid_online_config_no_delete_processed_data) {
+    ResetFlagsForTesting();
+    std::vector<std::string> args{
```
tessavitabile, line 357: My understanding is that after the flag is changed to a bool, it will be required, so I think the flag should be included here.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -764,6 +778,21 @@ void ParseAndValidateArguments(const vector<string> &args,
...
+
+    if (slamService.offlineFlag) {
```
tessavitabile, line 785: Same here, I think you can error if delete_processed_data is true and offlineFlag is false, since rdk should not pass this combination.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -880,6 +909,14 @@ int FindFrameIndex(const std::vector<std::string> &filesRGB,
     return -1;
 }
 
+int RemoveFile(std::string file_path) {
```
tessavitabile, line 912: The return can be void, since its value is not checked.

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
tessavitabile: You can change to void, both here and in cartographer.

slam-libraries/viam-orb-slam3/orbslam_server_v1_test.cc
```diff
@@ -100,6 +104,7 @@ BOOST_AUTO_TEST_CASE(ParseAndValidateArguments_valid_config) {
     BOOST_TEST(slamService.map_rate_sec.count() == chrono::seconds(60).count());
     BOOST_TEST(slamService.camera_name == "color");
     BOOST_TEST(slamService.offlineFlag == false);
+    BOOST_TEST(slamService.delete_processed_data == false);
 }
 
 BOOST_AUTO_TEST_CASE(
```
tessavitabile, line 110: For this test case, I would include the delete_processed_data parameter, since it should always be sent by rdk.

tessavitabile (approved):
LGTM mod comments

JohnN193 (approved):
LGTM, thanks for updating these to get tests to pass!

slam-libraries/viam-cartographer/src/slam_service/config.cc
jeremyhyde-viam: Yes I've added JIRA Ticket: https://viam.atlassian.net/browse/RSDK-1593 to cover this once other PRs are merged

slam-libraries/viam-cartographer/src/slam_service/config_test.cc
kkufieta: I think breaking this out is better and should be done in the same way in the orbslam tests.

slam-libraries/viam-orb-slam3/orbslam_server_v1_test.cc
```diff
@@ -100,26 +126,37 @@ BOOST_AUTO_TEST_CASE(ParseAndValidateArguments_valid_config) {
...
+                              "-sensors=",
+                              "-data_rate_ms=200",
+                              "-map_rate_sec=60",
+                              "-delete_processed_data=false"};
     SLAMServiceImpl slamService;
     utils::ParseAndValidateArguments(args, slamService);
     BOOST_TEST(slamService.camera_name == "");
     BOOST_TEST(slamService.offlineFlag == true);
+    BOOST_TEST(slamService.delete_processed_data == false);
 }
```
kkufieta, line 160: [minor] Instead of incorporating the `delete_processed_data` tests indirectly in other tests, break these out into a separate test case where you check for all four combinations of situations: offline/online vs. delete_processed_data=false/true.

kkufieta (changes requested):
First pass - a few nits, questions, and a minor.

slam-libraries/viam-cartographer/src/slam_service/config.cc
```diff
@@ -94,6 +104,23 @@ void ParseAndValidateConfigParams(int argc, char** argv,
     slamService.data_rate_ms = std::chrono::milliseconds(FLAGS_data_rate_ms);
     slamService.map_rate_sec = std::chrono::seconds(FLAGS_map_rate_sec);
 
+    // TODO: Remove once PR #1689 is submitted
+    // https://github.com/viamrobotics/rdk/pull/1689 This will allow intergation
+    // tests to pass (See associated JIRA ticket:
+    // https://viam.atlassian.net/browse/RSDK-1593)
+    if (FLAGS_delete_processed_data.empty()) {
+        slamService.delete_processed_data = false;
```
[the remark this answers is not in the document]
JohnN193: since the current integration tests are using a version of rdk that does not pass the `delete_process_data` flag into cartographer, this if statement combined with the string flag you asked about above is used to effectively ignore the missing input, and will be removed once integration tests pass in rdk.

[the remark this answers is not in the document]
kkufieta: Ok I think I understand - thanks for explaining!

slam-libraries/viam-cartographer/src/io/file_handler.cc
```diff
@@ -80,12 +80,11 @@ std::vector<std::string> ListSortedFilesInDirectory(
...
 
-int RemoveFile(std::string file_path) {
+void RemoveFile(std::string file_path) {
     if (remove(file_path.c_str()) != 0) {
-        LOG(INFO) << "Error removing file";
-        return 0;
+        LOG(ERROR) << "Error removing file";
     }
-    return 1;
+    return;
```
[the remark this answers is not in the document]
jeremyhyde-viam: As tess pointed out in their comment above, the success/failure is not being used anywhere so we can change it to return void. 

We could also add in checks but since most operations would want to continue even if it could remove a file the log here is enough (alternatively we could change the log to return an error if desirable)

slam-libraries/viam-cartographer/src/io/file_handler_test.cc
```diff
@@ -76,9 +76,7 @@ BOOST_AUTO_TEST_CASE(RemoveFile_success) {
 
     // Remove a file
     int file_num = 1;
-    int success =
-        RemoveFile(tmp_dir.string() + "/data/" + data_files.at(file_num));
-    BOOST_TEST(success == file_num);
```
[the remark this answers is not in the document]
jeremyhyde-viam: As per my comment above, we were turning this function into returning void. Since we check the contents of the folder below to confirm the correct file was removed, this check here was not necessary as a failure below would catch it

slam-libraries/viam-cartographer/src/slam_service/config.cc
```diff
@@ -94,6 +104,23 @@ void ParseAndValidateConfigParams(int argc, char** argv,
...
+    } else {
+        slamService.delete_processed_data =
+            !(FLAGS_delete_processed_data == "false");
+    }
```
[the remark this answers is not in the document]
jeremyhyde-viam: I was torn on this one because its a string so I need to check spelling of the word and choose to prioritize the true case (such that if a user wrote 'fales' it would trigger true). However since RDK will always spell true/false correctly I agree with you we can stick to the more legible version

slam-libraries/viam-cartographer/src/slam_service/config.cc
```diff
@@ -30,6 +30,14 @@ DEFINE_int64(
     "Frequency at which we want to print map pictures while cartographer "
     "is running.");
 DEFINE_string(input_file_pattern, "", "Input file pattern");
+// TODO: Switch back to DEFINE_bool once PR #1689 is submitted
+// https://github.com/viamrobotics/rdk/pull/1689 This will allow intergation
+// tests to pass (See associated JIRA ticket:
+// https://viam.atlassian.net/browse/RSDK-1593)
+DEFINE_string(delete_processed_data, "",
```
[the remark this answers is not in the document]
jeremyhyde-viam: because a bool doesn't have a nil type like string and as John said in his avoid comment, we are capitalizing on the fact that RDK always will send a `delete_processed_data` value to ensure the integration tests for this and the RDK PR. 

After the other PR is merged we could keep this as a string to be consistent with orbslam but i think it makes more sense to move it to the bool form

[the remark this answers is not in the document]
jeremyhyde-viam: Update: I've pushed a revision that uses the bool flag as, now that delete_processed_data is required in SLAM (still optional in RDK) we can use the default false value of the bool to not delete data (which is what the current tests want).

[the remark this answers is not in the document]
jeremyhyde-viam: Updates to the orbslam TODOs have been made as well which should reduce the change needed in follow up PR

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
tessavitabile: Ah, sorry, I didn't notice we use the return value in tests. If we do continue to have a return here, bool would be more idiomatic for modern C++

slam-libraries/viam-cartographer/src/io/file_handler.cc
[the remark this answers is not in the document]
tessavitabile: Agree that a bool would be better if we continue return a value. I agree that failing to remove a file shouldn't throw. Logging the error sounds good.

slam-libraries/viam-cartographer/src/slam_service/config.cc
```diff
@@ -30,6 +30,14 @@ DEFINE_int64(
...
+              "Deletes data after its been processed");
```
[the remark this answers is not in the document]
kkufieta: My bad - you're right. That's also correct!

kkufieta (approved):
LGTM - Thanks for making the changes I suggested!

PR #121, opened 2022-12-19
[RSDK-1039] only use installed lua files

JohnN193:
ticket removes the check from cartographer to look for lua files in the` <DATA_DIR>/config/lua_files` directory. In addition adds a make target for installing the lua files locally for dev use

https://viam.atlassian.net/browse/RSDK-1039


.github/workflows/main.yml
```diff
@@ -18,7 +18,7 @@ on:
 
 jobs:
   appimage:
-    uses: viamrobotics/slam/.github/workflows/appimage.yml@main
+    uses: viamrobotics/slam/.github/workflows/appimage.yml@RSDK-1039
```
JohnN193, line 21: will revert before merge

.github/workflows/pullrequest.yml
```diff
@@ -16,6 +16,6 @@ on:
 
 jobs:
   test:
-    uses: viamrobotics/slam/.github/workflows/test.yml@main
+    uses: viamrobotics/slam/.github/workflows/test.yml@RSDK-1039
```
JohnN193, line 19: will revert before merge

slam-libraries/Makefile
```diff
@@ -85,10 +85,14 @@ else
...
+install-lua-carto:
+	sudo mkdir -p /usr/local/share/cartographer/lua_files/
+	sudo cp viam-cartographer/lua_files/* /usr/local/share/cartographer/lua_files/
+	sudo cp viam-cartographer/cartographer/configuration_files/* /usr/local/share/cartographer/lua_files/
 
 testcarto:
 	cd viam-cartographer && ./scripts/test_cartographer.sh
 
-carto-all: bufinstall buf setupcarto buildcarto testcarto
+carto-all: bufinstall buf setupcarto buildcarto install-lua-carto testcarto
```
jeremyhyde-viam, line 96: [opt] I think install-lua-carto should be a part of the setup or build makes.

slam-libraries/viam-cartographer/src/slam_service/config.cc
```diff
@@ -69,23 +69,18 @@ void ParseAndValidateConfigParams(int argc, char** argv,
...
-        if (exists(relativePathToLuas)) {
-            LOG(INFO) << "Using lua files from relative path";
-            slamService.configuration_directory = relativePathToLuas.string();
-        } else {
-            LOG(ERROR) << "No lua files found, looked in "
-                       << pathToLuasFromConfig << " and " << relativePathToLuas;
-        }
+        LOG(ERROR) << "No lua files found, looked in " << relativePathToLuas;
+        LOG(ERROR)
+            << "Use 'make install-lua-carto' to install lua files on system";
```
tessavitabile, line 83: We should throw here, since this should terminate the program.

jeremyhyde-viam, line 83: [nit] ... install lua files into system's shared folder (or something like that)

tessavitabile (approved):
LGTM mod comment

jeremyhyde-viam (approved):
LGTM, one or two very minor comments

slam-libraries/viam-cartographer/src/slam_service/config.cc
JohnN193: only issue here is currently cartographer tests are intended to run without lua files being present. Since we checking for lua files relative to the binary location, the only way for the tests to locate lua files would be to install the tests to /usr/local/bin before using them, which results in the [tests passing](https://github.com/viamrobotics/slam/actions/runs/3747573312/jobs/6363944760#step:8:11) in CI (as of [this commit](https://github.com/viamrobotics/slam/pull/121/commits/115e6c567bc1e9234791a25105d1ff5708cc4cdd)). 

While this works for CI, it can make the dev experience clunky. To solve this i could possibly add an install target for the cartographer unit tests and add that to `make carto-all`. Should I go ahead and do that & add the throw here? Or is it better to avoid installing more stuff on a user's computer?

also worth noting that if lua files are not found, cartographer does throw because of them at a later step

tessavitabile: Thanks for explaining. I think it's fine if cartographer throws later.

kkufieta (approved):
LGTM

kkufieta (commented):
Actually I forgot - I think the README(s) still needs to be updated?

slam-libraries/Makefile
JohnN193: wanted it separate from setup just because its a unique dependency that installs locally rather than using a package manager, and we do not want our build steps to install things, users should actively choose to install the dependency.

slam-libraries/viam-cartographer/README.md
```diff
@@ -89,10 +89,11 @@ sudo cp build/carto_grpc_server /usr/local/bin/
...
 ## Running cartographer
 
-In your desired data directory, move the configuration files from cartographer into your `~/config` folder:  
+In your desired data directory, move the configuration files from cartographer into /usr/local/share`:  
 ```bash
-cp -r lua_files ~/YOUR_DATA_DIR/config/
-cp cartographer/configuration_files/* ~/YOUR_DATA_DIR/config/lua_files/
+sudo mkdir -p /usr/local/share/cartographer/lua_files
+cp lua_files/* /usr/local/share/cartographer/lua_files/
+cp cartographer/configuration_files/* /usr/local/share/cartographer/lua_files/
```
kkufieta, line 96: [q] Why can one skip `sudo` here but not in the Makefile?

kkufieta (approved):
Thanks for fixing the README's! 

LGTM - just one curiosity

slam-libraries/viam-cartographer/README.md
JohnN193: totally needs sudo

PR #122, opened 2022-12-19
RSDK-1048 Fix Reading Time from Timestamp

jeremyhyde-viam:
Fixed parsing of timestamp and failures in testing due to not checking std::istringstream failure mode.

JIRA Ticket: [RSDK-1048](https://viam.atlassian.net/browse/RSDK-1048)


JohnN193 (approved):
LGTM, good catch on the partial time format

kkufieta (approved):
Great catch! LGTM

PR #123, opened 2022-12-27
[RSDK-1044] Adjust log levels

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-1044

* `VLOG(1)` is used for debug-level logging. Disabled by default (`v = 0`). Enable by setting `v = 1`.
* `LOG(...)` is used for info, warning, error, and fatal logging. It is not advised to change the setting for this since some of the functionalities (e.g. port forwarding) rely on obtaining the logs.

More information: https://github.com/google/glog#setting-flags

### DIFF
--- a/slam-libraries/viam-cartographer/src/main.cc
+++ b/slam-libraries/viam-cartographer/src/main.cc
@@ -47,7 +47,7 @@ int main(int argc, char** argv) {
     std::unique_ptr<grpc::Server> server(builder.BuildAndStart());
 
     // This log line is needed by rdk to get the port.
-    LOG(INFO) << "Server listening on " << *selected_port << "\n";
+    VLOG(0) << "Server listening on " << *selected_port << "\n";
 
     slamService.RunSLAM();
 
--- a/slam-libraries/viam-cartographer/src/main.cc
+++ b/slam-libraries/viam-cartographer/src/main.cc
@@ -11,7 +11,7 @@
 #include "slam_service/slam_service.h"
 
 void exit_loop_handler(int s) {
-    LOG(INFO) << "Finishing session.\n";
+    LOG(INFO) << "Finishing session.";
     viam::b_continue_session = false;
 }
 
--- a/slam-libraries/viam-cartographer/src/main.cc
+++ b/slam-libraries/viam-cartographer/src/main.cc
@@ -17,7 +17,7 @@ void exit_loop_handler(int s) {
 
 int main(int argc, char** argv) {
     // glog only supports logging to files and stderr, not stdout.
-    FLAGS_alsologtostderr = 1;
+    FLAGS_logtostderr = 1;
     google::InitGoogleLogging(argv[0]);
     struct sigaction sigIntHandler;
 
--- a/slam-libraries/viam-cartographer/src/main.cc
+++ b/slam-libraries/viam-cartographer/src/main.cc
@@ -47,7 +47,7 @@ int main(int argc, char** argv) {
     std::unique_ptr<grpc::Server> server(builder.BuildAndStart());
 
     // This log line is needed by rdk to get the port.
-    VLOG(0) << "Server listening on " << *selected_port << "\n";
+    LOG(INFO) << "Server listening on " << *selected_port << "\n";
 
     slamService.RunSLAM();
 
--- a/slam-libraries/viam-cartographer/src/mapping/map_builder.cc
+++ b/slam-libraries/viam-cartographer/src/mapping/map_builder.cc
@@ -67,7 +67,7 @@ void MapBuilder::LoadMapFromFile(std::string map_filename,
         map_builder_->pose_graph()->RunFinalOptimization();
     }
     for (auto&& trajectory_ids_pair : trajectory_ids_map)
-        LOG(INFO) << "Trajectory ids mapping from apriori map: "
+        VLOG(1) << "Trajectory ids mapping from apriori map: "
                   << trajectory_ids_pair.first << " "
                   << trajectory_ids_pair.second;
 }
@@ -77,7 +77,7 @@ void MapBuilder::SaveMapToFile(bool include_unfinished_submaps,
     bool ok = map_builder_->SerializeStateToFile(include_unfinished_submaps,
                                                  filename_with_timestamp);
     if (!ok) {
-        LOG(WARNING) << "Saving the map to pbstream failed.";
+        LOG(ERROR) << "Saving the map to pbstream failed.";
     }
 }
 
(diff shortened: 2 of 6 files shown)


slam-libraries/viam-cartographer/src/main.cc
```diff
@@ -17,7 +17,7 @@ void exit_loop_handler(int s) {
 
 int main(int argc, char** argv) {
     // glog only supports logging to files and stderr, not stdout.
-    FLAGS_alsologtostderr = 1;
+    FLAGS_logtostderr = 1;
```
kkufieta, line 20: Changed this to log in stderr only - as opposed to both the log file and stderr. 

[q] Should we keep logging both to the log file and to stderr? Or should we keep this change?

jeremyhyde-viam (commented):
To clarify:

Without debug = 1:
We see LOG(ERROR) and LOG(FATAL)
With debug = 0
We see VLOG(1) LOG(ERROR) and LOG(FATAL)

slam-libraries/viam-cartographer/src/main.cc
jeremyhyde-viam: I haven't used the log files yet so unless you have I think we can remove

slam-libraries/viam-cartographer/src/main.cc
```diff
@@ -11,13 +11,13 @@
 #include "slam_service/slam_service.h"
 
 void exit_loop_handler(int s) {
-    LOG(INFO) << "Finishing session.\n";
+    LOG(INFO) << "Finishing session.";
```
jeremyhyde-viam, line 14: I might switch this and the system shutdown log to VLOG(0) so that users always see that (or change it to VLOG(1)

slam-libraries/viam-cartographer/src/mapping/map_builder.cc
```diff
@@ -67,17 +67,17 @@ void MapBuilder::LoadMapFromFile(std::string map_filename,
         map_builder_->pose_graph()->RunFinalOptimization();
     }
     for (auto&& trajectory_ids_pair : trajectory_ids_map)
-        LOG(INFO) << "Trajectory ids mapping from apriori map: "
-                  << trajectory_ids_pair.first << " "
-                  << trajectory_ids_pair.second;
+        VLOG(1) << "Trajectory ids mapping from apriori map: "
```
jeremyhyde-viam, line 70: If you agree with my initial comment this should be LOG I think

slam-libraries/viam-cartographer/src/mapping/map_builder.cc
```diff
@@ -67,17 +67,17 @@ void MapBuilder::LoadMapFromFile(std::string map_filename,
...
+                << trajectory_ids_pair.second;
 }
 
 void MapBuilder::SaveMapToFile(bool include_unfinished_submaps,
                                const std::string filename_with_timestamp) {
     bool ok = map_builder_->SerializeStateToFile(include_unfinished_submaps,
                                                  filename_with_timestamp);
     if (!ok) {
-        LOG(WARNING) << "Saving the map to pbstream failed.";
+        LOG(ERROR) << "Saving the map to pbstream failed.";
```
jeremyhyde-viam, line 80: good switch

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -538,7 +538,7 @@ void SLAMServiceImpl::ProcessDataAndStartSavingMaps(double data_start_time) {
         // Set TrajectoryBuilder
         trajectory_id = map_builder.SetTrajectoryBuilder(&trajectory_builder,
                                                          {kRangeSensorId});
-        LOG(INFO) << "Using trajectory ID: " << trajectory_id;
+        VLOG(1) << "Using trajectory ID: " << trajectory_id;
```
jeremyhyde-viam, line 541: [opt] see previous comment

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -538,7 +538,7 @@ void SLAMServiceImpl::ProcessDataAndStartSavingMaps(double data_start_time) {
...
     }
 
     LOG(INFO) << "Beginning to add data...";
```
jeremyhyde-viam, line 544: [opt] would switch to VLOG based on review comment

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -621,8 +621,8 @@ void SLAMServiceImpl::ProcessDataAndStartSavingMaps(double data_start_time) {
             optimization_lock.lock();
 
             std::lock_guard<std::mutex> lk(map_builder_mutex);
-            LOG(INFO)
-                << "Starting to optimize final map. Can take a little while...";
+            LOG(INFO) << "Starting to optimize final map. This can take a "
```
jeremyhyde-viam, line 624: [opt] VLOG

jeremyhyde-viam (commented):
In my opinion we should use VLOG(X) to represent slam service logs and LOG(X) to represent internal cartographer ones and file processing. I.e. VLOGs should be higher level and LOGs more technical. Does this align with your thoughts?

slam-libraries/viam-cartographer/src/main.cc
kkufieta: yeah I tried that but it didn't show up in the log messages which is why I left it as is.

kkufieta:
> In my opinion we should use VLOG(X) to represent slam service logs and LOG(X) to represent internal cartographer ones and file processing. I.e. VLOGs should be higher level and LOGs more technical. Does this align with your thoughts?

@jeremyhyde-viam I tried to do that initially caused that's recommended. I then quickly found out that other libraries we're importing are using VLOG(1) excessively - which means that even if we wanted to define our own VLOG scale for viam-cartographer and leave LOGs for native cartographer - we wouldn't be able to see our VLOG lines due to excessive spamming from our dependencies.

There might be a way to deal with this using different VLOG settings for modules, but that seemed out of scope for what this ticket tried to do (a quick fix to improve the current situation if it doesn't take up too much time).

slam-libraries/viam-cartographer/src/main.cc
kkufieta: I haven't either - I think I'd like to hear if @tessavitabile  has any objections to this change

jeremyhyde-viam (commented):
I understand, I think a complete overhaul of the logging can be done during the refactoring into modules. 

For now I think its most important that for non-debug testing (minimal/default logs) and CI/CD (desired logs) to be correct level of verbosity.

After looking at your tests, I think the level is correct in CI/CD. I will look into the local test as well to confirm minimal/default logs.

jeremyhyde-viam (approved):
LGTM, more work will be done during modularization

slam-libraries/viam-cartographer/src/main.cc
tessavitabile: No objection

PR #128, opened 2023-01-03
[RSDK-1255] fix slam process exit handler

JohnN193:
Currently the slam algorithms fail to shut down when rdk attempts to close them. This PR attempts to update the exit handler to safely close out of the slam algorithms, based off information found in[ this PR](https://github.com/viamrobotics/goutils/pull/116). Will have a followup PR in rdk to clean up the integration test code

Note the changes make the clean shutdown occur only when running the slam servers with rdk. Running the binaries standalone/without managed process controlling them is not a behavior we need to support. 

https://viam.atlassian.net/browse/RSDK-1255

[RSDK-1255]: https://viam.atlassian.net/browse/RSDK-1255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/slam-libraries/viam-cartographer/src/main.cc
+++ b/slam-libraries/viam-cartographer/src/main.cc
@@ -19,13 +19,13 @@ int main(int argc, char** argv) {
     // glog only supports logging to files and stderr, not stdout.
     FLAGS_logtostderr = 1;
     google::InitGoogleLogging(argv[0]);
-    struct sigaction sigIntHandler;
+    struct sigaction sigTermHandler;
 
-    sigIntHandler.sa_handler = exit_loop_handler;
-    sigemptyset(&sigIntHandler.sa_mask);
-    sigIntHandler.sa_flags = 0;
+    sigTermHandler.sa_handler = exit_loop_handler;
+    sigemptyset(&sigTermHandler.sa_mask);
+    sigTermHandler.sa_flags = 0;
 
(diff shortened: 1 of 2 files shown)


JohnN193:
We now see the[ exit handler triggers ](https://github.com/viamrobotics/slam/actions/runs/3832288046/jobs/6522424589#step:12:370)from the slam repos, so we know managed process is closing out of the slam algorithms. We also no longer have the [unable to check exit status logs](https://github.com/viamrobotics/rdk/actions/runs/3687572584/jobs/6241321866#step:6:474) which we saw in the past, which indicated the issue was present

Otterverse (commented):
So I'm glad that this is working now, but... is there any foreseeable situation where people run these binaries directly? If so, you should probably handle both SIGINT and SIGTERM (you can use the same handler) so that a human user can still use CTRL+C to exit the program gracefully, rather than having that be an uncaught signal and/or do nothing. If that's just NEVER going to be a concern though, then it's probably fine as is.

kkufieta (commented):
I second James's concerns - I think no matter the likelihood of someone running the binaries directly, we should support clean shutdown for both situations.

Otterverse (approved):
LGTM!

kkufieta (approved):
LGTM

PR #133, opened 2023-01-09
RSDK-1073 Add GetPositionNew to Cartographer

jeremyhyde-viam:
This PR adds the `GetPositionNew` endpoint to cartographer. This `GetPositionNew` will eventually replace the current `GetPosition`. It differs from `GetPosition` in two of its return parameters:

- Instead of returning a `PoseInFrame` it now returns a `Pose`
- There is an additional string return (`component_reference`) that represents the component that should be used to connect the SLAM outputs to RDK's proprioceptive reference frame system.

JIRA Ticket: [RSDK-1073](https://viam.atlassian.net/browse/RSDK-1073)

Note: Tested using dev branch in cartographer integration tests. Will add tests in once the dev parameter PR is submitted and we are ready to start testing endpoints.

[RSDK-1073]: https://viam.atlassian.net/browse/RSDK-1073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -52,6 +52,43 @@ std::atomic<bool> b_continue_session{true};
     return grpc::Status::OK;
 }
 
+::grpc::Status SLAMServiceImpl::GetPositionNew(
+    ServerContext *context, const GetPositionNewRequest *request,
+    GetPositionNewResponse *response) {
+    cartographer::transform::Rigid3d global_pose;
+    {
+        std::lock_guard<std::mutex> lk(viam_response_mutex);
+        global_pose = latest_global_pose;
+    }
+
+    // Set pose for our response
+    Pose *myPose = response->mutable_pose();
+    myPose->set_x(global_pose.translation().x());
+    myPose->set_y(global_pose.translation().y());
+    myPose->set_z(global_pose.translation().z());
+
+    // Set component_reference for our response
+    // auto *myComponentReference = response->mutable_component_reference();
+    // myComponentReference->set__component_reference(camera_name)
+
+    response->set_component_reference(camera_name)
+
+        // Set extra for our response (currently stores quaternion)
+        google::protobuf::Struct *q;
+    google::protobuf::Struct *extra = response->mutable_extra();
+    q = extra->mutable_fields()->operator[]("quat").mutable_struct_value();
+    q->mutable_fields()->operator[]("real").set_number_value(
+        global_pose.rotation().w());
+    q->mutable_fields()->operator[]("imag").set_number_value(
+        global_pose.rotation().x());
+    q->mutable_fields()->operator[]("jmag").set_number_value(
+        global_pose.rotation().y());
+    q->mutable_fields()->operator[]("kmag").set_number_value(
+        global_pose.rotation().z());
+
+    return grpc::Status::OK;
+}
+
 ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
                                        const GetMapRequest *request,
                                        GetMapResponse *response) {
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -68,9 +68,6 @@ std::atomic<bool> b_continue_session{true};
     myPose->set_z(global_pose.translation().z());
 
     // Set component_reference for our response
-    // auto *myComponentReference = response->mutable_component_reference();
-    // myComponentReference->set__component_reference(camera_name)
-
     response->set_component_reference(camera_name);
 
     // Set extra for our response (currently stores quaternion)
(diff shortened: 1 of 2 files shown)


JohnN193 (approved):
LGTM

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -52,6 +52,42 @@ ::grpc::Status SLAMServiceImpl::GetPosition(ServerContext *context,
     return grpc::Status::OK;
 }
 
+// TODO 01/10/2023: Remove current GetPosition and rename this implementation to
```
tessavitabile, line 55: [nit] No need for this TODO, since it won't be hard to find this when implementing the ticket. If you do prefer to keep the TODO, I would write this as:
TODO RSDK-1066: Remove...

tessavitabile (approved):

kkufieta (approved):

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
jeremyhyde-viam: Removing

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -52,6 +52,40 @@ ::grpc::Status SLAMServiceImpl::GetPosition(ServerContext *context,
     return grpc::Status::OK;
 }
 
+::grpc::Status SLAMServiceImpl::GetPositionNew(
+    ServerContext *context, const GetPositionNewRequest *request,
```
nicksanford, line 56: I know this is all the same as the original, but in the interest of me learning:

why is `GetPositionNewRequest *request` `const`?

nicksanford: never mind, I figured it out

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -52,6 +52,40 @@ ::grpc::Status SLAMServiceImpl::GetPosition(ServerContext *context,
...
+    GetPositionNewResponse *response) {
+    cartographer::transform::Rigid3d global_pose;
+    {
+        std::lock_guard<std::mutex> lk(viam_response_mutex);
```
nicksanford, line 60: where is `lk` defined? 
I don't see it either here: https://cplusplus.com/reference/mutex/mutex/ or here: https://cplusplus.com/reference/mutex/lock_guard/

I assume it is the same as `lock` but I want to know where it is defined.

nicksanford: Also, what unlocks the mutex?

JohnN193:
> > 
> 
> I assume the answer is `no`, but for my own edification: is it not possible to test any of these changes outside of an integration test and if so why?

I actually believe we could tests these outside of integration tests technically, we would just need to spin up a client similar to how tests in `services/slam/server_test.go` happen. But since the slam libraries are only intended to work with rdk it would be out of scope to set those tests up

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -52,6 +52,40 @@ ::grpc::Status SLAMServiceImpl::GetPosition(ServerContext *context,
...
+        global_pose.rotation().w());
+    q->mutable_fields()->operator[]("imag").set_number_value(
+        global_pose.rotation().x());
+    q->mutable_fields()->operator[]("jmag").set_number_value(
+        global_pose.rotation().y());
+    q->mutable_fields()->operator[]("kmag").set_number_value(
+        global_pose.rotation().z());
+
+    // Set component_reference for our response
+    response->set_component_reference(camera_name);
```
nicksanford, line 84: Where is `set_component_reference` defined?

nicksanford (approved):
Approved as the code appears to be the exact same as existing code except for a single function. 

The questions are just for my own understanding.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: the mutex gets unlocked when we leave the scope the lock is defined for, so we create the lock, copy our values into a temp variable(here `global_pose`), then release the lock on `latest_global_pose` as we have exited its scope

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: `set_component_reference` is defined in our [slam proto](https://github.com/viamrobotics/api/blob/main/proto/viam/service/slam/v1/slam.proto#L104) and takes the `camera_name` from the [sensors](https://github.com/viamrobotics/slam/blob/main/slam-libraries/viam-cartographer/src/slam_service/config.cc#L25) input. Also def worth noting that if sensors are not included b/c of leaving it out of the config in offline mode, then an empty string will be set for the cartographer case

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
nicksanford: got it, the scope defines the duration of the lock.

Thanks for the explanation, still ramping up in C++!

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
jeremyhyde-viam: It is a grpc definition from proto

PR #139, opened 2023-01-25
[RSDK-1062] implement orb GetPointCloudMap

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-1062)

Add GetPointCloudMap GRPC endpoint to orbslam.

Tested via:

```bash
# request new endpoint
grpcurl -d '{"name": "run-slam"}'  -plaintext -protoset <(~/slam-clone/slam-libraries/grpc/bin/buf build -o -) $(sudo netstat -tulpn | grep LISTEN | grep orb_grpc_se | awk '{print $4}' | sort | head -n 1) viam.service.slam.v1.SLAMService/GetPointCloudMap | jq -r '.pointCloudPcd' | base64 -d > get_point_cloud_map.pcd

# request old endpoint
grpcurl -d '{"name": "run-slam", "mime_type": "pointcloud/pcd"}'  -plaintext -protoset <(~/slam-clone/slam-libraries/grpc/bin/buf build -o -) $(sudo netstat -tulpn | grep LISTEN | grep orb_grpc_ | awk '{print $4}' | sort | head -n 1) viam.service.slam.v1.SLAMService/GetMap | jq -r '.pointCloud.pointCloud' | base64 -d > get_map.pcd

#convert new & old response value to ascii for easier analysis
pcl_convert_pcd_ascii_binary get_map.pcd get_map_ascii3.pcd 0
pcl_convert_pcd_ascii_binary get_point_cloud_map.pcd get_point_cloud_map_ascii3.pcd 0

# observe they are the same except for the color
vimdiff get_map_ascii3.pcd get_point_cloud_map_ascii3.pcd
```
[get_point_cloud_map_ascii3.pcd.txt](https://github.com/viamrobotics/slam/files/10520635/get_point_cloud_map_ascii3.pcd.txt)
[get_map_ascii3.pcd.txt](https://github.com/viamrobotics/slam/files/10520641/get_map_ascii3.pcd.txt)

 attached

I also confirmed that app.viam UI is able to render the new returned point cloud:

With the `dev` config feature flag off (i.e. old GetMap endpoint):
old

With the `dev` config feature flag on (i.e. old GetPointCloudMap endpoint):
new

### DIFF
--- a/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
+++ b/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
@@ -140,8 +140,10 @@ iterates throught all the bytes of the float
 writes each byte to buffer 
 */
 void writeFloatToBufferInBytes(std::string buffer, float f) {
+    BOOST_LOG_TRIVIAL(debug) << "writing float: " << f << "\nsize of float: " << sizeof(float);
     unsigned char const * const p = (unsigned char const *)(&f);
     for (std::size_t i = 0; i < sizeof(float); ++i) {
+        BOOST_LOG_TRIVIAL(debug) << "index " << i << " as integer: " << p[i] << "\n";
         buffer.push_back(p[i]);
     }
 }
--- a/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
+++ b/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
@@ -116,11 +116,9 @@ std::atomic<bool> b_continue_session{true};
     float max = 0;
     float min = 10000;
     std::vector<ORB_SLAM3::MapPoint *> actualMap;
-    Sophus::SE3f currPose;
     {
         std::lock_guard<std::mutex> lk(slam_mutex);
         actualMap = currMapPoints;
-        currPose = poseGrpc;
     }
 
     if (actualMap.size() == 0) {
--- a/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
+++ b/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
@@ -39,7 +39,6 @@ const auto HEADERTEMPLATE =
     "POINTS %d\n"
     "DATA binary\n";
 
-
 std::atomic<bool> b_continue_session{true};
 
 ::grpc::Status SLAMServiceImpl::GetPosition(ServerContext *context,
--- a/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
+++ b/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
@@ -24,21 +24,42 @@ using viam::common::v1::PoseInFrame;
 #define MAX_COLOR_VALUE 255
 const std::string strRGB = "/rgb";
 const std::string strDepth = "/depth";
-const std::string HEADERTEMPLATE = "VERSION .7\n"
-                                   "FIELDS x y z\n"
-                                   // NOTE: If a float is more than 4 bytes
-                                   // on a given platform
-                                   // this size will be inaccurate
-                                   "SIZE 4 4 4\n"
-                                   "TYPE F F F\n"
-                                   "COUNT 1 1 1\n"
-                                   "WIDTH %d\n"
-                                   "HEIGHT 1\n"
-                                   "VIEWPOINT 0 0 0 1 0 0 0\n"
-                                   "POINTS %d\n"
-                                   "DATA binary\n";
+const std::string HEADERTEMPLATE =
+    "VERSION .7\n"
+    "FIELDS x y z\n"
+    // NOTE: If a float is more than 4 bytes
+    // on a given platform
+    // this size will be inaccurate
+    "SIZE 4 4 4\n"
+    "TYPE F F F\n"
+    "COUNT 1 1 1\n"
+    "WIDTH %d\n"
+    "HEIGHT 1\n"
+    "VIEWPOINT 0 0 0 1 0 0 0\n"
+    "POINTS %d\n"
+    "DATA binary\n";
 
 namespace viam {
+
+/*
+applies the mapSize to the HEADERTEMPLATE
+returning the pcd header as a string
+*/
+std::string pcdHeader(int mapSize) {
+    return str(boost::format(HEADERTEMPLATE) % mapSize % mapSize);
+}
+
+/*
+casts the float f to a pointer of unsigned bytes
+iterates throught all the bytes of f
+writes each byte to buffer
+*/
+void writeFloatToBufferInBytes(std::string *buffer, float f) {
+    unsigned const char *const p = (unsigned const char *)(&f);
+    for (std::size_t i = 0; i < sizeof(float); ++i) {
+        buffer->push_back(p[i]);
+    }
+}
 
 std::atomic<bool> b_continue_session{true};
 
@@ -125,39 +146,6 @@ std::atomic<bool> b_continue_session{true};
     return grpc::Status::OK;
 }
 
-
-/*
-applies the mapSize to the HEADERTEMPLATE
-returning the pcd header as a string
-*/
-std::string pcdHeader(int mapSize) {
-    return str(boost::format(HEADERTEMPLATE) % mapSize % mapSize);
-}
-
-/*
-casts the float f to a pointer of unsigned bytes
-iterates throught all the bytes of f
-writes each byte to buffer 
-*/
-void writeFloatToBufferInBytes(std::string* buffer, float f) {
-    unsigned const char * const p = (unsigned const char *)(&f);
-    for (std::size_t i = 0; i < sizeof(float); ++i) {
-        buffer->push_back(p[i]);
-    }
-}
-
-/*
-casts the uint32_t f to a pointer of unsigned bytes
-iterates throught all the bytes of f
-writes each byte to buffer 
-*/
-void writeIntUnsignedToBufferInBytes(std::string* buffer, uint32_t f) {
-    unsigned const char * const p = (unsigned const char *)(&f);
-    for (std::size_t i = 0; i < sizeof(uint32_t); ++i) {
-        buffer->push_back(p[i]);
-    }
-}
-
 ::grpc::Status SLAMServiceImpl::GetPointCloudMap(
     ServerContext *context, const GetPointCloudMapRequest *request,
     GetPointCloudMapResponse *response) {
@@ -168,7 +156,8 @@ void writeIntUnsignedToBufferInBytes(std::string* buffer, uint32_t f) {
     }
 
     if (actualMap.size() == 0) {
-        return grpc::Status(grpc::StatusCode::UNAVAILABLE, "currently no map points exist");
+        return grpc::Status(grpc::StatusCode::UNAVAILABLE,
+                            "currently no map points exist");
     }
 
     // Take sparse slam map and convert into a pcd.
--- a/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
+++ b/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
@@ -140,7 +140,7 @@ iterates throught all the bytes of the float
 writes each byte to buffer 
 */
 void writeFloatToBufferInBytes(std::string* buffer, float f) {
-    const char * const p = (const char *)(&f);
+    unsigned const char * const p = (unsigned const char *)(&f);
     for (std::size_t i = 0; i < sizeof(float); ++i) {
         buffer->push_back(p[i]);
     }
--- a/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
+++ b/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
@@ -136,12 +136,24 @@ std::string pcdHeader(int mapSize) {
 
 /*
 casts the float f to a pointer of unsigned bytes
-iterates throught all the bytes of the float
+iterates throught all the bytes of f
 writes each byte to buffer 
 */
 void writeFloatToBufferInBytes(std::string* buffer, float f) {
     unsigned const char * const p = (unsigned const char *)(&f);
     for (std::size_t i = 0; i < sizeof(float); ++i) {
+        buffer->push_back(p[i]);
+    }
+}
+
+/*
+casts the uint32_t f to a pointer of unsigned bytes
+iterates throught all the bytes of f
+writes each byte to buffer 
+*/
+void writeIntUnsignedToBufferInBytes(std::string* buffer, uint32_t f) {
+    unsigned const char * const p = (unsigned const char *)(&f);
+    for (std::size_t i = 0; i < sizeof(uint32_t); ++i) {
         buffer->push_back(p[i]);
     }
 }
@@ -198,7 +210,7 @@ void writeFloatToBufferInBytes(std::string* buffer, float f) {
         cv::cvtColor(hsv, valRGB2, cv::COLOR_HSV2RGB);
         cv::Vec3b colorRGB = valRGB2.at<cv::Vec3b>(cv::Point(0, 0));
 
-        int rgb = 0;
+        uint32_t rgb = 0;
         rgb = rgb | ((int)colorRGB[0] << 16);
         rgb = rgb | ((int)colorRGB[1] << 8);
         rgb = rgb | ((int)colorRGB[2] << 0);
@@ -206,7 +218,7 @@ void writeFloatToBufferInBytes(std::string* buffer, float f) {
         writeFloatToBufferInBytes(&buffer, v.x());
         writeFloatToBufferInBytes(&buffer, v.y());
         writeFloatToBufferInBytes(&buffer, v.z());
-        writeFloatToBufferInBytes(&buffer, rgb);
+        writeIntUnsignedToBufferInBytes(&buffer, rgb);
     }
     response->set_point_cloud_pcd(buffer);
     return grpc::Status::OK;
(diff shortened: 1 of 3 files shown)


slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -110,6 +110,86 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
     return grpc::Status::OK;
 }
 
+::grpc::Status SLAMServiceImpl::GetPointCloudMap(
```
nicksanford, line 113: This is equal to this if statement https://github.com/viamrobotics/slam/pull/139/files#diff-b68d96da1bc970406f0f4402811d582812550e7c8450a636d73540f0363c73d8L318, just fitting it into a `GetPointCloudMapResponse` return value.

nicksanford:
I've been told by the team that these types of GRPC methods are not tested by unit tests & therefor are not automatically tested in the slam repo itself. 

I'm curious if people would find it valuable to add integration / GRPC test to this repo? We could maybe use a python, go or any language with a GRPC client the team would find ergonomic to write tests in. Doing so would automate the manual testing I have been doing with grpcurl to verify the endpoints behave how I expect.

I'd be happy to take on that work if we would find it valuable. Also completely happy if the answer is "post modularization we can consider it".

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -110,6 +110,86 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
...
+    ServerContext *context, const GetPointCloudMapRequest *request,
+    GetPointCloudMapResponse *response) {
+    float max = 0;
+    float min = 10000;
+    std::vector<ORB_SLAM3::MapPoint *> actualMap;
+    Sophus::SE3f currPose;
```
tessavitabile, line 119: [nit] no need to get currPose

tessavitabile (approved):
LGTM mod nit

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -110,6 +110,84 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
...
+    if (actualMap.size() == 0) {
+        return grpc::Status(grpc::StatusCode::UNAVAILABLE,
+                            "currently no map points exist");
+    }
+
+    // take sparse slam map and convert into a pcd. Orientation of PCD
+    // is wrt the camera (z is coming out of the lens) so may need to
+    // transform.
+
+    std::stringbuf buffer;
```
nicksanford, line 133: @JohnN193 I copied the pattern from here https://github.com/viamrobotics/slam/pull/9/files#diff-b68d96da1bc970406f0f4402811d582812550e7c8450a636d73540f0363c73d8R136 in `GetMap` & have a few questions:


1. When the original GetMap code was introduced james provided some feedback that this could be refactored to use sprintf, would this now be a good time to do that refactor?
2. Why are we creating both a `std::stringbuf` , and then creating a `std::ostream` from it & then writing to both the ostream in some parts of the code ([here](https://github.com/viamrobotics/slam/pull/139/files#diff-b68d96da1bc970406f0f4402811d582812550e7c8450a636d73540f0363c73d8R137) & to the stringbuf [here](https://github.com/viamrobotics/slam/pull/139/files#diff-b68d96da1bc970406f0f4402811d582812550e7c8450a636d73540f0363c73d8R182) in others? Ideally I'd like to only have a single variable which is written to, but please educate me if we actually need to have 2 different objects with pointers to each other for some reason. 


cc @tessavitabile for C++ expertise

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -110,6 +110,84 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
...
+    // "ground". Map written as a binary
+    for (auto p : actualMap) {
+        Eigen::Matrix<float, 3, 1> v = p->GetWorldPos();
+        float val = v.y();
+        auto ratio = (val - mid) / span;
+
+        clr = (int)(offsetRGB + (ratio * spanRGB));
+
+        cv::Vec3b &color = hsv.at<cv::Vec3b>(cv::Point(0, 0));
+        color[0] = clr;
```
nicksanford, line 174: After talking with @JohnN193 it turns out that I need to remove color information from this given that orbslam doesn't have probability information & the RGB channel is currently arbitrarily representing hight.

I'm also going to try to refactor this code to have helper functions for the ascii & binary writing to the buffer & try to only use a single buffer variable for the entire function.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
jeremyhyde-viam: I was looking into ifstream and simialr functions while doing GetInternalState [carto] and while it may not always be the case, ostream takes in a buffer object (either a file buffer  or string buffer). To my knowledge the ostream is handling the addition of bytes to the buffer object.

You might be able to get away with just a stringbuf but the ostream helps optimize performance.

https://stackoverflow.com/questions/48728330/whats-the-mechanism-of-iostreams-with-buffer

https://cplusplus.com/reference/ostream/ostream/ostream/

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -110,6 +110,84 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
     return grpc::Status::OK;
 }
 
+::grpc::Status SLAMServiceImpl::GetPointCloudMap(
+    ServerContext *context, const GetPointCloudMapRequest *request,
+    GetPointCloudMapResponse *response) {
+    float max = 0;
+    float min = 10000;
```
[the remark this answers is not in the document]
jeremyhyde-viam: Looks like they are used on line 150 to help calculate the color values. You can just change this to MaxInt and MinInt if you want to remove magic numbers but use the same color calc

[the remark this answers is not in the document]
JohnN193: talked on this offline, these were for scaling the current height to color which is something we do not need/want anymore

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
JohnN193: yeah refactoring here would be fine, should help clean stuff up

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
tessavitabile: oh right, I forgot how we're using colors currently!

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -110,6 +110,84 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
...
+    int clr = 0;
+    cv::Mat hsv(1, 1, CV_8UC3, cv::Scalar(255, 255, 255));
+    cv::Mat valRGB2(hsv.size(), hsv.type());
+
+    // write the map with simple rgb colors based off height from the
+    // "ground". Map written as a binary
+    for (auto p : actualMap) {
+        Eigen::Matrix<float, 3, 1> v = p->GetWorldPos();
+        float val = v.y();
+        auto ratio = (val - mid) / span;
```
jeremyhyde-viam, line 169: [minor/nit] can we split all of the color math to its own function? I'd love to have GetMap be as streamlined as possible for ease of reference in the future

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -39,6 +41,10 @@ class SLAMServiceImpl final : public SLAMService::Service {
     ::grpc::Status GetMap(ServerContext *context, const GetMapRequest *request,
                           GetMapResponse *response) override;
 
+    ::grpc::Status GetPointCloudMap(
```
jeremyhyde-viam, line 44: Can we add comments to at the very least the GRPC functions?

jeremyhyde-viam (changes requested):
Good job moving this over, glad it isnt looking as involved as cartographer's GetPointCloudMap! A few minor improvements and answers to posed questions

tessavitabile:
By the way, just a reminder to use the imperative mood for commit messages, and make the PR title the same as the intended commit message: https://viam.atlassian.net/wiki/spaces/ENG/pages/262371/Viam+Engineering+Guidelines

nicksanford:
> 

Thanks for the reminder!

Once this is all approved I'll squash all changes & rename the commit message to the title. I believe that will bring this PR back into compliance with your reminder but please let me know if I've missed something. 

Also, if I'm squashing on merge, do the incremental commits (which happen during debugging / responding to feedback) require adhering to the imperative mood (given that they will be squashed away on merge I believe)?

tessavitabile:
No, incremental commits don't require imperative mood. Only the commit that gets merged.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
nicksanford: + a refactor to remove the color information (as it is not longer meaningful) & refactor to only use a single string type to create the pointcloud pcd, with helper functions where possible.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -22,9 +25,43 @@ using viam::common::v1::PoseInFrame;
...
+std::string pcdHeader(int mapSize) {
+    return str(boost::format(HEADERTEMPLATE) % mapSize % mapSize);
+}
+
+/*
+casts the float f to a pointer of unsigned 8 bit bytes
+iterates throught all the 8 bit bytes of f
+writes each 8 bit bytes to buffer
+*/
+void writeFloatToBufferInBytes(std::string &buffer, float f) {
```
nicksanford, line 58: I'm passing buffer as a reference as I'm under the impression (from the tour of C++ book) that passing references rather than pointers is preferable whenever possible, which it is in this case.  Please let me know if that is not the case.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -22,9 +25,43 @@ using viam::common::v1::PoseInFrame;
...
+    auto p = (const uint8_t *)(&f);
+    for (std::size_t i = 0; i < sizeof(float); ++i) {
```
nicksanford, line 60: @tessavitabile could I trouble you to confirm this is the safest way to get each of the 8 bytes out of a 32 bit float & add each byte to a string? I use a `uint8_t` pointer as I'm under the impression `const char *` (the pointer type the old implementation cast to) may not always be 8 bits on all platforms, which would break the expectations of our implementation, but please correct me if there is an error in my thinking / a different way I should do this.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
nicksanford: done

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
nicksanford: done

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
nicksanford: team decided to remove color info all together, so I think this is now addressed.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
tessavitabile: I've also seen the pattern of passing a const reference (for things that can be const) and a non-const pointer (for things that cannot be const). I'm not sure why people use pointers for non-const sometimes. But I think what you've done here makes sense.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
tessavitabile: I'm not sure this is necessary here. We only support a couple platforms, and we test those platforms. If we were to strive for platform independence, I would want to test on more platforms (and we would also have other things to fix, such as assuming the size of a float in the pcd header). For now, I think we shouldn't make parts of the code platform-independent, if it adds complexity.

tessavitabile (changes requested):
I think it's cool that you looked into this, but I don't recommend making the change for platform independence at this time.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
tessavitabile: We discussed offline that the motivation is readability. I don't feel that I can speak to the safest way to perform this--I would be looking it up for myself.

One suggestion I have is to add a check that sizeof(float)==4 and fail fast if this is not the case. Otherwise, LGTM

tessavitabile (approved):

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
jeremyhyde-viam: For reference, Cartographer serialization uses this function:
```
void WriteSizeAsLittleEndian(uint64 size, std::ostream* out) {
  for (int i = 0; i != 8; ++i) {
    out->put(size & 0xff);
    size >>= 8;
  }
}
```
just as a reference but as Tess already pointed out I think we can focus for now on a single platform and your function can be expanded upon once we deal with more platforms.

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -39,6 +41,19 @@ class SLAMServiceImpl final : public SLAMService::Service {
     ::grpc::Status GetMap(ServerContext *context, const GetMapRequest *request,
                           GetMapResponse *response) override;
 
+    /*
+      For a given GetPointCloudMapRequest
+      Returns a GetPointCloudMapResponse containing a sparse
+      slam map as a Binary PCD file
```
jeremyhyde-viam, line 47: [nit] remove file as PCD is a valid format to refer to a buffer and having file here might make people expect a saved file

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -39,6 +41,19 @@ class SLAMServiceImpl final : public SLAMService::Service {
...
+
+      Map uses:
+      1. right hand rule
```
jeremyhyde-viam, line 50: [q] not sure if its worth stating that we use the right hand rule as the left hand rule is only ever used when looking at magnetic fields but clarity is never a bad thing

jeremyhyde-viam (approved):
LGTM, one or two nits but otherwise should do perfectly. Thanks for providing testing instructions too

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -22,9 +25,43 @@ using viam::common::v1::PoseInFrame;
 #define MAX_COLOR_VALUE 255
 const std::string strRGB = "/rgb";
 const std::string strDepth = "/depth";
+const auto HEADERTEMPLATE =
```
JohnN193, line 28: [q] should this be defined within the viam::utils namespace? Or is it okay to leave as a global

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -22,9 +25,43 @@ using viam::common::v1::PoseInFrame;
...
+    "VIEWPOINT 0 0 0 1 0 0 0\n"
+    "POINTS %d\n"
+    "DATA binary\n";
 
 namespace viam {
 
+/*
+applies the mapSize to the HEADERTEMPLATE
+returning the pcd header as a string
+*/
```
JohnN193, line 48: [opt]we haven't been using block comments for other functions, not something that needs to change though

JohnN193 (approved):
Spoke offline about moving the helper functions into the `utils` namespace. otherwise just some small stuff, LGTM

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
nicksanford: Thanks for the review y'all. Just to get another opinion I asked eric & he said that this is pretty much the only way to get each byte out of a float in C++ (which is not super safe no matter what one does) & is the same in terms of safety to the original.

As a result I'll keep this as it is a win for readability.

He did bring up 2, longer term (maybe for us to chat about later @tessavitabile) concerns, not with this implementation but more with the [PCD spec](https://pointclouds.org/documentation/tutorials/pcd_file_format.html) itself) which says:
`in binary form, where the data is a complete memory copy of the pcl::PointCloud.points array/vector. On Linux systems, we use mmap/munmap operations for the fastest possible read/write access to the data.`

What that means is that data corruption will likely silently occur if either of the following are true:
1.  the encoder & decoder processors have different endianness (i.e. not little endian)
2. the encoder & decoder processors have different floating point representations (i.e. not ieee 754)

Currently all our code assumes little endian (as that is what amd & intel support) & ieee 754.

This isn't relevant to this change but we may want to consider providing additional metadata in the PCD response regarding the floating point spec to use to interpret the bytes & what endianness to use as well, by checking on the PCD encoder side for these behaviors on boot.

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
nicksanford: done

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
nicksanford: done

PR #135, opened 2023-01-13
RSDK-1061 [orb] GetPositionNew

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-1061)

Implements viam.service.slam.v1.SLAMService/GetPositionNew on orbslamv3 slam library.

How to see the output:
```bash
grpcurl -d '{"name": "slam"}'  -plaintext -protoset <(~/slam-clone/slam-libraries/grpc/bin/buf build -o -) $(sudo netstat -tulpn | grep LISTEN | grep orb_grpc_serv | awk '{print $4}' | sort | head -n 1) viam.service.slam.v1.SLAMService/GetPositionNew
{
  "pose": {
    "x": 0.007205881178379059,
    "y": -0.004640873987227678,
    "z": 0.021260758861899376
  },
  "componentReference": "color",
  "extra": {
      "quat": {
            "imag": 0.009775514714419842,
            "jmag": 0.006308149080723524,
            "kmag": -0.010251480154693127,
            "real": 0.999879777431488
          }
    }
}
```

PR #140, opened 2023-01-26
merge-bufcompile-into-buf

nicksanford:
Revert edit to Makefile https://github.com/viamrobotics/slam/pull/135/files#diff-931a6e2149fb04d2ec1e4ae28596d823e5671314e365f0a590f6959d792caf80L11

### DIFF
--- a/slam-libraries/Makefile
+++ b/slam-libraries/Makefile
@@ -8,9 +8,7 @@ bufsetup:
 	ln -sf `which grpc_cpp_plugin` grpc/bin/protoc-gen-grpc-cpp
 	pkg-config openssl || export PKG_CONFIG_PATH=$$PKG_CONFIG_PATH:`find \`which brew > /dev/null && brew --prefix\` -name openssl.pc | head -n1 | xargs dirname`
 
-buf: bufsetup bufcompile
-
-bufcompile:
+buf: bufsetup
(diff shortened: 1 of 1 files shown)


Otterverse:
Talked with John on this, found the issue. The new "buf" target now depends on two other targets, whereas previously it depended on one, then its own commands ran after that. Now it tries to run both at the same time, and the buf install doesn't finish before it tries to use it.

Basically "bufcompile: bufsetup" will fix it properly. Reverting it, as you are doing, is fine too.

Otterverse:
Note that you can reproduce this in a normal environment with just:

```
rm grpc/bin/buf
make -j4 buf
```

Otterverse (approved):
Please see other comments about WHY this is happening, but this reversion is fine too.

nicksanford:
> 

Thanks @JohnN193 & @Otterverse for helping diagnose. Yep, that makes sense. I'm going to merge this as is given that making buf depend on  bufcompile and bufcompile depend bufsetup ends up meaning that the buf command will do the exact same work as bufcompile, which defeats the purpose of having 2 commands. 

In the figure, I'll be sure to ensure that when defining targets that depend on more than one command don't care about the execution order of the dependent commands.

JohnN193 (approved):

PR #141, opened 2023-01-29
[RSDK-1816] update submodules to use http links instead of ssh 

JohnN193:
submodule links made the reppo unable to retrieve the submodules unless there was an ssh key in use with github. In addition the ssh links would cause the brew packages to fail on systems without an ssh key.

ticket: https://viam.atlassian.net/browse/RSDK-1816


Otterverse (approved):
LGTM

PR #142, opened 2023-01-31
Rsdk 1063 implement orb GetInternalState

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-1063)

NOTE: Depending on approval of the scope doc to make GetInternalState & GetPointCloudMap server streaming endpoints, this PR may need a follow-up PR to refactor into a server streaming API before considering the ticket complete. 

Tested with:
```bash
grpcurl -max-msg-sz 320000000  -d '{"name": "run-slam"}'  -plaintext -protoset <(~/slam-clone/slam-libraries/grpc/bin/buf build -o -) $(sudo netstat -tulpn | grep LISTEN | grep orb_grpc_serv | awk '{print $4}' | sort | head -n 1) viam.service.slam.v1.SLAMService/GetInternalState | jq -r '.internalState' | base64 -d > orb_internal_state.osa
```

Depends on https://github.com/viamrobotics/ORB_SLAM3/pull/7 to add `DumpOsa` to `ORB_SLAM3::System`

Testing steps (from @JohnN193  thank you!):
```
Localization Mode. 
This assumes localization mode will return the state of the map, even before localization has occurred.

This verifies that all the slam state needed to run Orbslam was contained within the GetInternalState response.

1. Do an initial run with no initial map. The config should use live data. Data rate_msec should be set to 200
2. GetPointCloudMap & persist to PCD file, GetInternalState & persist to file
3. Stop RDK
4. Move the file to the maps directory, with the same name as the most recently written map file with the current timestamp.
5. Start RDK
6. Confirm you see `Initialization of Atlas from file` with the file you copied into the map directory.
7. Confirm Orbslam does not crash & you are able to get a successful response from GetPointCloudMap
```

I ran through the testing steps above this PR passes them.

### DIFF
--- a/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
+++ b/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
@@ -428,8 +428,49 @@ std::atomic<bool> b_continue_session{true};
 ::grpc::Status SLAMServiceImpl::GetInternalState(ServerContext *context,
                                        const GetInternalStateRequest *request,
                                        GetInternalStateResponse *response) {
-                                        
-    return grpc::Status::OK;
+
+    std::stringbuf buffer;
+    bool success = ArchiveSlam(buffer);
+    if (success) {
+        response->set_internal_state(buffer.str());
+        return grpc::Status::OK;
+    } else {
+        return grpc::Status(grpc::StatusCode::UNAVAILABLE,
+                            "SLAM is not yet initialized");
+    }
+}
+
+// TODO: This setter existing is an antipattern,
+// which only exists b/c we only have one class
+// for both the data thread(s) & GRPC server & the
+// fact that it takes a while to fully initialize
+// the slam algo (longer than we want to wait to 
+// boot the GRPC server, so as to not hit timeouts
+// in RDK).
+// In the future there should be a class that initializes
+// the slam object when it is initialized
+// so that the slam pointer can never be null.
+// It only exists so that ArchiveSlam (called by the
+// GRPC handlers) can get access to the slam object
+// to call DumpOsa.
+void SLAMServiceImpl::SetSlam(ORB_SLAM3::System* s) {
+    slam = s;
+}
+
+bool SLAMServiceImpl::ArchiveSlam(std::stringbuf &buffer) {
+    // I belive this is the right place to put this mutex as
+    // if I put it as the first line within the if statement,
+    // there is no gurantee slam hasn't been set to nullptr
+    // by another thread immediately after the if statement
+    // predicate completes but before the mutex starts.
+    std::lock_guard<std::mutex> lk(slam_mutex);
+    if (slam == nullptr) {
+        BOOST_LOG_TRIVIAL(debug) << "ArchiveSlam slam is NULL";
+        return false;
+    }
+    BOOST_LOG_TRIVIAL(debug) << "ArchiveSlam slam is NOT NULL";
+    slam->DumpOsa(buffer);
+    return true;
 }
 
 void SLAMServiceImpl::ProcessDataOnline(ORB_SLAM3::System *SLAM) {
--- a/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
+++ b/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
@@ -425,6 +425,13 @@ std::atomic<bool> b_continue_session{true};
     return grpc::Status::OK;
 }
 
+::grpc::Status SLAMServiceImpl::GetInternalState(ServerContext *context,
+                                       const GetInternalStateRequest *request,
+                                       GetInternalStateResponse *response) {
+                                        
+    return grpc::Status::OK;
+}
+
 void SLAMServiceImpl::ProcessDataOnline(ORB_SLAM3::System *SLAM) {
     std::vector<std::string> filesRGB = utils::ListFilesInDirectoryForCamera(
         path_to_data + strRGB, ".png", camera_name);
--- a/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
+++ b/slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
@@ -449,7 +449,10 @@ std::atomic<bool> b_continue_session{true};
 // so that the SLAM pointer can never be null.
 // SetSlam only exists so that ArchiveSlam will have access
 // to the SLAM object when called by the GRPC handlers.
-void SLAMServiceImpl::SetSlam(ORB_SLAM3::System *s) { slam = s; }
+void SLAMServiceImpl::SetSlam(ORB_SLAM3::System *s) {
+    std::lock_guard<std::mutex> lk(slam_mutex);
+    slam = s;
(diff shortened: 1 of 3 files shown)


slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -425,6 +425,54 @@ ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
...
+
+bool SLAMServiceImpl::ArchiveSlam(std::stringbuf &buffer) {
+    // I belive this is the right place to put this mutex as
+    // if I put it as the first line within the if statement,
+    // there is no gurantee slam hasn't been set to nullptr
+    // by another thread immediately after the if statement
+    // predicate completes but before the mutex starts.
+    std::lock_guard<std::mutex> lk(slam_mutex);
+    if (slam == nullptr) {
+        BOOST_LOG_TRIVIAL(debug) << "ArchiveSlam slam is NULL";
```
nicksanford, line 468: I remove the call to SetSlam in main & verified that this null check behaves as expected if `ArchiveSlam` is called before `slam` is set.

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -94,6 +104,7 @@ class SLAMServiceImpl final : public SLAMService::Service {
     std::atomic<bool> finished_processing_offline{false};
     std::thread *thread_save_atlas_as_osa_with_timestamp;
 
+    ORB_SLAM3::System *slam;
```
nicksanford, line 107: Please let me know if I should be using unique pointer or a different data type for this.

slam-libraries/viam-orb-slam3/orbslam_server_v1_main.cc
```diff
@@ -170,6 +169,7 @@ int main(int argc, char **argv) {
         SLAM->GetAtlas()->ChangeMap(largestMap);
     }
 
+    slamService.SetSlam(SLAM.get());
```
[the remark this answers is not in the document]
JohnN193: i saw this, ty for the cleanup!

nicksanford:
I will add integration tests to RDK once this PR is merged & won't complete the ticket until integration tests are in RDK.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -425,6 +425,50 @@ ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
...
+// to call DumpOsa.
+void SLAMServiceImpl::SetSlam(ORB_SLAM3::System *s) { slam = s; }
+
+bool SLAMServiceImpl::ArchiveSlam(std::stringbuf &buffer) {
+    // I belive this is the right place to put this mutex as
+    // if I put it as the first line within the if statement,
+    // there is no gurantee slam hasn't been set to nullptr
+    // by another thread immediately after the if statement
+    // predicate completes but before the mutex starts.
+    std::lock_guard<std::mutex> lk(slam_mutex);
```
nicksanford, line 463: I believe this is the right place to put this mutex as if I put it as the first line within the if statement, there is no guarantee slam hasn't been set to nullptr by another thread immediately after the if statement predicate completes but before the mutex starts.

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
JohnN193: I think we would want a unique pointer if possible, as that is the idiomatic(am I using that word right) way to handle this for c++

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -52,6 +55,10 @@ class SLAMServiceImpl final : public SLAMService::Service {
         ServerContext *context, const GetPointCloudMapRequest *request,
         GetPointCloudMapResponse *response) override;
 
+    ::grpc::Status GetInternalState(
```
tessavitabile, line 58: [nit] Can we be consistent with whether we are commenting the gRPC methods (either add comments to GetInternalState and GetPositionNew, or remove comment from GetPointCloudMap)?

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
tessavitabile: We can only have one unique pointer per raw pointer, so you should either leave this as a raw pointer or move ownership of the raw pointer here in main().

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
tessavitabile: We also need to hold the mutex while setting `slam` for this to be true

slam-libraries/viam-orb-slam3/orbslam_server_v1_main.cc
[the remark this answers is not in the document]
tessavitabile: We should also call `slamService.SetSlam(nullptr)` prior to the end of this function. Otherwise, if `SLAM` gets destroyed before `server`, a call to `GetInternalState` may try to access invalid memory.

tessavitabile (changes requested):

JohnN193:
> I will add integration tests to RDK once this PR is merged & won't complete the ticket until integration tests are in RDK.

The new integration test will not pass in RDK unless this PR is merged first. I think testing the integration test locally before merging here is fine.

JohnN193 (approved):
LGTM, but make sure the orbslam module is updated before merging

slam-libraries/viam-orb-slam3/orbslam_server_v1_main.cc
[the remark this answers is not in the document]
nicksanford: I've added it. Please let me know if that is sufficient to make it safe.

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
nicksanford: done

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
nicksanford: Ok, mutex acquired within `SetSlam`

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
nicksanford: Got it. I spent an hour trying out refactoring the code to make it so that there is only one unique pointer to slam & using that pointer both for the data threads & for the ArchiveSlam method however, it ended up being quite a large change (one which I would love to do) but I get the impression we are trying to defer any large refactors until modularization.

Is leaving this as is sufficiently safe?

tessavitabile (approved):

nicksanford:
@tessavitabile thank you very much for taking the time to teach me about unique pointers during this PR review!

PR #144, opened 2023-02-06
[RSDK-1074] Implement GetPointCloudMap for cartographer

JohnN193:
Implements GetPointCloudMap in cartographer. tested using the `dev: true` attribute in the slam service config with an offline dataset. 

Additionally causes Cartographer to shut down if an error occurs when trying to generate a painted map. This error was previously swallowed and now shuts down Cartographer with `std::terminate()`. 
JIRA ticket: [RSDK-1074](https://viam.atlassian.net/browse/RSDK-1074?atlOrigin=eyJpIjoiNDZkM2Y1ZjEzNjJlNDRmYThmZWJhZDg5YWNhYTEyM2IiLCJwIjoiaiJ9)

[RSDK-1074]: https://viam.atlassian.net/browse/RSDK-1074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -131,9 +131,6 @@ std::atomic<bool> b_continue_session{true};
         oss << "error writing pointcloud to response " << e.what();
         return grpc::Status(grpc::StatusCode::UNAVAILABLE, oss.str());
     }
-    // return grpc::Status::OK;
-    return grpc::Status(grpc::StatusCode::UNAVAILABLE,
-                        "currently working on map endpoint");
 }
 
 ::grpc::Status SLAMServiceImpl::GetMap(ServerContext *context,
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -30,6 +30,7 @@ std::atomic<bool> b_continue_session{true};
         std::lock_guard<std::mutex> lk(viam_response_mutex);
         global_pose = latest_global_pose;
     }
+
     // Setup mapping of pose message to the response. NOTE not using
     // inFrame->set_reference_frame
     PoseInFrame *inFrame = response->mutable_pose();
@@ -64,24 +65,26 @@ std::atomic<bool> b_continue_session{true};
         global_pose = latest_global_pose;
     }
 
+    // Rotate pose to XZ plane. Additional angle offset is used so rotations
+    // occur along the Y axis, to match the XZ plane
+    auto rotated_vector = pcdRotation * global_pose.translation();
+    auto rotated_quat =
+        pcdOffsetRotation * global_pose.rotation() * pcdRotation;
+
     // Set pose for our response
     Pose *myPose = response->mutable_pose();
-    myPose->set_x(global_pose.translation().x());
-    myPose->set_y(global_pose.translation().y());
-    myPose->set_z(global_pose.translation().z());
+    myPose->set_x(rotated_vector.x());
+    myPose->set_y(rotated_vector.y());
+    myPose->set_z(rotated_vector.z());
 
     // Set extra for our response (currently stores quaternion)
     google::protobuf::Struct *q;
     google::protobuf::Struct *extra = response->mutable_extra();
     q = extra->mutable_fields()->operator[]("quat").mutable_struct_value();
-    q->mutable_fields()->operator[]("real").set_number_value(
-        global_pose.rotation().w());
-    q->mutable_fields()->operator[]("imag").set_number_value(
-        global_pose.rotation().x());
-    q->mutable_fields()->operator[]("jmag").set_number_value(
-        global_pose.rotation().y());
-    q->mutable_fields()->operator[]("kmag").set_number_value(
-        global_pose.rotation().z());
+    q->mutable_fields()->operator[]("real").set_number_value(rotated_quat.w());
+    q->mutable_fields()->operator[]("imag").set_number_value(rotated_quat.x());
+    q->mutable_fields()->operator[]("jmag").set_number_value(rotated_quat.y());
+    q->mutable_fields()->operator[]("kmag").set_number_value(rotated_quat.z());
 
     // Set component_reference for our response
     response->set_component_reference(camera_name);
@@ -164,7 +167,6 @@ std::atomic<bool> b_continue_session{true};
 
 ::grpc::Status SLAMServiceImpl::GetCurrentPointCloudMap(const GetMapRequest *request,
                                                  GetMapResponse *response) {
-    bool pointcloud_has_points = false;
     std::string pointcloud_map;
     // Write or grab the latest pointcloud map in form of a string
     try {
@@ -174,12 +176,11 @@ std::atomic<bool> b_continue_session{true};
             // We are able to lock the optimization_shared_mutex, which means
             // that the optimization is not ongoing and we can grab the newest
             // map
-            pointcloud_has_points = ExtractPointCloudToString(pointcloud_map);
+            GetLatestSampledPointCloudMapString(pointcloud_map);
         } else {
             // We couldn't lock the mutex which means the optimization process
             // locked it and we need to use the backed up latest map
             std::lock_guard<std::mutex> lk(viam_response_mutex);
-            pointcloud_has_points = latest_pointcloud_map_has_points;
             pointcloud_map = latest_pointcloud_map;
         }
     } catch (std::exception &e) {
@@ -191,7 +192,7 @@ std::atomic<bool> b_continue_session{true};
     // Write the pointcloud map string to the response
     try {
         common::v1::PointCloudObject *pco = response->mutable_point_cloud();
-        if (!pointcloud_has_points) {
+        if (pointcloud_map.empty()) {
             LOG(ERROR) << "map pointcloud does not have points yet";
             return grpc::Status(grpc::StatusCode::UNAVAILABLE,
                                 "map pointcloud does not have points yet");
@@ -277,13 +278,11 @@ void SLAMServiceImpl::BackupLatestMap() {
     std::string jpeg_map_with_marker_tmp = GetLatestJpegMapString(true);
     std::string jpeg_map_without_marker_tmp = GetLatestJpegMapString(false);
     std::string pointcloud_map_tmp;
-    bool pointcloud_map_has_points_tmp =
-        ExtractPointCloudToString(pointcloud_map_tmp);
+    GetLatestSampledPointCloudMapString(pointcloud_map_tmp);
 
     std::lock_guard<std::mutex> lk(viam_response_mutex);
     latest_jpeg_map_with_marker = std::move(jpeg_map_with_marker_tmp);
     latest_jpeg_map_without_marker = std::move(jpeg_map_without_marker_tmp);
-    latest_pointcloud_map_has_points = pointcloud_map_has_points_tmp;
     latest_pointcloud_map = std::move(pointcloud_map_tmp);
 }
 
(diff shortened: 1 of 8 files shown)


slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -92,6 +92,52 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
     return grpc::Status::OK;
 }
 
+::grpc::Status SLAMServiceImpl::GetPointCloudMap(
+    ServerContext *context, const GetPointCloudMapRequest *request,
+    GetPointCloudMapResponse *response) {
+    std::string pointcloud_map;
+    // Write or grab the latest pointcloud map in form of a string
+    try {
```
JohnN193, line 100: chose to copy/paste the now-named `GetCurrentPointCloudMap` instead of using that function because the request/response messages are different, and The previous version will no longer be needed once we remove the depreciated endpoints

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -92,6 +92,52 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
...
+        } else {
+            // We couldn't lock the mutex which means the optimization process
+            // locked it and we need to use the backed up latest map
+            std::lock_guard<std::mutex> lk(viam_response_mutex);
+            pointcloud_map = latest_pointcloud_map;
+        }
+    } catch (std::exception &e) {
+        LOG(ERROR) << "error creating pcd map: " << e.what();
+        std::ostringstream oss;
+        viam::b_continue_session = false;
```
JohnN193, line 117: Setting `b_continue_session` here. Currently, this is also used in [main.cc](https://github.com/viamrobotics/slam/blob/main/slam-libraries/viam-cartographer/src/main.cc#L15). I think atomic bool is supposed to be safe but I am going from memory

tessavitabile: Yes, it's safe to concurrently access an atomic bool. However, I don't think we should the normal shutdown procedure here, since the state is corrupted. I think you should just re-throw the exception and let that terminate the process, so we won't do any additional work. Alternatively, if you're worried about callers catching the exception, you can use std::terminate.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -142,7 +188,8 @@ ::grpc::Status SLAMServiceImpl::GetJpegMap(const GetMapRequest *request,
         }
     } catch (std::exception &e) {
         std::ostringstream oss;
-        oss << "error encoding image " << e.what();
+        oss << "error encoding image: " << e.what();
+        viam::b_continue_session = false;
```
tessavitabile, line 192: Same here, I would just re-throw the exception.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -178,6 +225,7 @@ ::grpc::Status SLAMServiceImpl::GetPointCloudMap(const GetMapRequest *request,
     } catch (std::exception &e) {
         std::ostringstream oss;
         oss << "error encoding pointcloud " << e.what();
+        viam::b_continue_session = false;
```
tessavitabile, line 228: Same here

tessavitabile (approved):
LGTM mod comments

Does the map in the UI look correct?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: `throw` was not working within the grpc functions for some reason, but std::terminate seems to work! Swapping to that

JohnN193:
The origin of the map is kat's desk from the old office. Original map(Pre PR)
Screen Shot 2023-02-07 at 12 23 46 PM

Current Map(2/7/2023 @12:30)
Screen Shot 2023-02-07 at 12 23 03 PM

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -92,6 +93,52 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
...
+        } else {
+            // We couldn't lock the mutex which means the optimization process
+            // locked it and we need to use the backed up latest map
+            std::lock_guard<std::mutex> lk(viam_response_mutex);
+            pointcloud_map = latest_pointcloud_map;
+        }
+    } catch (std::exception &e) {
+        LOG(ERROR) << "error creating pcd map: " << e.what();
+        std::ostringstream oss;
+        std::terminate();
```
tessavitabile, line 118: Nothing after the std::terminate call will be executed, so it should be removed.

tessavitabile (changes requested):

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -92,6 +93,50 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
...
+            GetLatestSampledPointCloudMapString(pointcloud_map);
+        } else {
+            // We couldn't lock the mutex which means the optimization process
+            // locked it and we need to use the backed up latest map
+            std::lock_guard<std::mutex> lk(viam_response_mutex);
+            pointcloud_map = latest_pointcloud_map;
+        }
+    } catch (std::exception &e) {
+        LOG(ERROR) << "error creating pcd map: " << e.what();
+        std::ostringstream oss;
```
tessavitabile, line 117: [nit] remove unused variable

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -142,8 +187,8 @@ ::grpc::Status SLAMServiceImpl::GetJpegMap(const GetMapRequest *request,
         }
     } catch (std::exception &e) {
         std::ostringstream oss;
```
tessavitabile, line 189: [nit] remove unused variable (or log this instead)

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -178,7 +223,7 @@ ::grpc::Status SLAMServiceImpl::GetPointCloudMap(const GetMapRequest *request,
...
```
tessavitabile, line 224: [nit] remove unused variable (or log this instead)

tessavitabile (approved):
LGTM mod comments

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -92,6 +93,50 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
...
+        response->set_point_cloud_pcd(pointcloud_map);
+        return grpc::Status::OK;
+    } catch (std::exception &e) {
+        std::ostringstream oss;
+        oss << "error writing pointcloud to response " << e.what();
+        return grpc::Status(grpc::StatusCode::UNAVAILABLE, oss.str());
+    }
+    // return grpc::Status::OK;
+    return grpc::Status(grpc::StatusCode::UNAVAILABLE,
+                        "currently working on map endpoint");
```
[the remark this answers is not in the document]
JohnN193: That was the return I used when initially testing the GetPointCloudMap endpoint, will remove

kkufieta (approved):

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
nicksanford: Can we add compiler warnings to fail the build if we have unused variables? Checking for this seems like something a compiler would be good at.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -92,6 +93,47 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
...
+                                           std::defer_lock};
+        if (optimization_lock.try_lock()) {
+            // We are able to lock the optimization_shared_mutex, which means
+            // that the optimization is not ongoing and we can grab the newest
+            // map
+            GetLatestSampledPointCloudMapString(pointcloud_map);
+        } else {
+            // We couldn't lock the mutex which means the optimization process
+            // locked it and we need to use the backed up latest map
+            std::lock_guard<std::mutex> lk(viam_response_mutex);
```
nicksanford, line 112: Do we want to log in each of these cases so if it turns out that we are frequently returning stale data we have some easy way to detect that?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -92,6 +93,47 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
...
+    std::string pointcloud_map;
+    // Write or grab the latest pointcloud map in form of a string
+    try {
+        std::shared_lock optimization_lock{optimization_shared_mutex,
+                                           std::defer_lock};
+        if (optimization_lock.try_lock()) {
+            // We are able to lock the optimization_shared_mutex, which means
+            // that the optimization is not ongoing and we can grab the newest
+            // map
+            GetLatestSampledPointCloudMapString(pointcloud_map);
```
nicksanford, line 108: Is there any reason why we don't set `latest_pointcloud_map` in this case?

If I'm understanding this code correctly, the way it is currently written its possible a caller to get cached data that is older than the last GetPointCloudMap call they performed.

That can happen if:
1. the cached data is from time T1
2. `GetPointCloudMap` is called & the `optimization_lock` is acquired. A map from time T2 is returned (but not cached)
3. `GetPointCloudMap`  is called & the `optimization_lock` is NOT acquired (as it is held by the data thread). The cached data is returned from T1 which is OLDER than the data the client just received. 

If I were a client, I would not be satisfied with sometimes getting a map older than the map I was just given. 

I would expect `GetPointCloudMap` to either:
1. Give me the map it just gave me
2. Give me a newer, better map


@tessavitabile please correct me if I'm mistaken about any of the above.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
nicksanford: Will this leak resources such as file handles, or will RAII ensure they are all closed?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
nicksanford: Also, as a note for the future (no change needed here): when writing functions that only exist to return a string (like this one does), we should stop writing functions which take a string, mutate the string & return void. Instead, we should just return the string by value.  

From reading the Tour of C++ book's section on strings, the creator of C++ recommends (section 10.2) returning strings by value, rather than mutating a string argument & relying on the caller to know that the argument now contains the data supplied by the function. Also, the efficiency of both approaches should be the same as frequently no copy is performed when returning a string by value, as std::sting implement the move constructor (and also b/c apparently, the compiler is able to use copy elusion to even avoid the move). 

@tessavitabile Please correct me if any of that is inaccurate (trying to get my C++ legs here).

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -92,6 +93,47 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
...
+        if (pointcloud_map.empty()) {
+            LOG(ERROR) << "map pointcloud does not have points yet";
+            return grpc::Status(grpc::StatusCode::UNAVAILABLE,
+                                "map pointcloud does not have points yet");
+        }
+        response->set_point_cloud_pcd(pointcloud_map);
+        return grpc::Status::OK;
+    } catch (std::exception &e) {
+        std::ostringstream oss;
+        oss << "error writing pointcloud to response " << e.what();
```
nicksanford, line 132: Should we log this condition so we have evidence know our end users are getting errors outside of the end user's web browser?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -421,11 +472,16 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
         // 2D SLAM so Z is set to 0
         float z_pos = 0;
+        Eigen::Vector3d map_point(x_pos, y_pos, z_pos);
+        auto rotated_map_point = pcdRotation * map_point;
 
         // Rotating coordinates to match slam service expectation (XZ plane)
-        viam::utils::writeFloatToBufferInBytes(data_buffer, x_pos);
-        viam::utils::writeFloatToBufferInBytes(data_buffer, z_pos);
-        viam::utils::writeFloatToBufferInBytes(data_buffer, y_pos);
+        viam::utils::writeFloatToBufferInBytes(data_buffer,
```
nicksanford, line 479: Is this fixing a bug in the original endpoint? If so, what was the bug & what is the new intended behavior?

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -51,11 +53,13 @@ static const int samplingFactor = 1;
 // Conversion to number of bytes used in colored PCD encoding
 static const int pixelBytetoPCDByte = 16 / 4;
 // Quaternion to rotate axes to the XZ plane
-static const Eigen::Quaterniond pcdRotation(0.7071068, 0.7071068, 0, 0);
+static const Eigen::Quaterniond pcdRotation(0.7071068, -0.7071068, 0, 0);
```
nicksanford, line 56: Why is this changing?

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -51,11 +53,13 @@ static const int samplingFactor = 1;
...
 // Static offset quaternion, so orientation matches physical intuition.
 // This will result in rotations occurring within the y axis to match 2D mapping
 // in the XZ plane
-static const Eigen::Quaterniond pcdOffsetRotation(0.7071068, -0.7071068, 0, 0);
+static const Eigen::Quaterniond pcdOffsetRotation(0.7071068, 0.7071068, 0, 0);
```
nicksanford, line 60: Why is this changing?

nicksanford (changes requested):

nicksanford:
@JohnN193 can you also show what you get when you run the pcd through this PCD viewer: 

You can get the PCD easily with:
```
# install grpcurl , make sure ~/go/bin is in your path
go install github.com/fullstorydev/grpcurl/cmd/grpcurl@latest

# from within the root directory of the api repo, assuming you have buf installed at `~/slam/slam-libraries/grpc/bin/buf` (if not just change that path to wherever buf is). Also if you are on mac rather than a pi just change `$(sudo netstat -tulpn | grep LISTEN | grep orb_grpc_ | awk '{print $4}' | sort | head -n 1) ` to whatever the host & port is of your cartogropher GRPC server. 

grpcurl -d '{"name": "run-slam"}'  -plaintext -protoset <(~/slam/slam-libraries/grpc/bin/buf build -o -) $(sudo netstat -tulpn | grep LISTEN | grep orb_grpc_ | awk '{print $4}' | sort | head -n 1) viam.service.slam.v1.SLAMService/GetPointCloudMap | jq -r '.pointCloudPcd' | base64 -d > get_point_cloud_map.pcd
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
JohnN193: the change here was to flip the map in the viewer on cartographer's side, to get the map in the expected orientation

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: new intended behavior is so if there are any other transformations needed for cartographer's endpoint that they get captured. Previous behavior had me manually rotating the cartographer map points(z_pos and y_pos being flipped). 

If we end up needing more changes to the map orientation I wanted to make it easier/clearer to capture them for both the Position and PointCloud endpoints. Can add a comment at the math in GetPointcloud so its clear that we are using the rotation here too

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: I think you're right that there is a race condition where optimization caches a map, then GetPointCloudMap creates and returns a new map, then GetPointCloudMap returns the cached map. However, no changes would occur to the map during this time because optimization only runs at the end of processing in offline mode (and also at the beginning when loading an a priori map, but GetPointCloudMap will not return anything at this time). So I don't think we'll actually see maps go backward in time.

Yes, it is better to return a string by value.

JohnN193: Tess beat me to it, but yeah this optimization lock should only be NOT acquired in offline mode at the end of a run or on startup if a flag is enabled. 

As far as setting `latest_pointcloud_map` is concerned I personally don't see the harm in it, so I'll set it up in the relevant functions

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
JohnN193: should have written at the top of the PR but additional fixes this PR looked to tackle are:

1. capturing the failed serialization state
2. a fix for the map appearing "flipped" when viewing, though I am only fixing this for the PCD endpoint.

JohnN193:
Here is the map I have been using in the other PCD viewer. As for artifacts, I currently cannot get the dataset to download
Screen Shot 2023-02-09 at 1 20 02 PM

JohnN193:
Using dataset 15 from the fake slam PR

Screen Shot 2023-02-09 at 2 34 43 PM
Screen Shot 2023-02-09 at 2 28 58 PM

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
nicksanford: cool, thanks for explaining! I missed the optimization step nuance.

nicksanford (approved):
Looks great as do the new maps!!! 🗺️

PR #145, opened 2023-02-09
[RSDK-1785] do not update map when in localization mode

JohnN193:
Adds a cached copy of the point cloud map when entering localization mode. Was not implemented for the jpeg endpoint as that will be removed shortly. Tested using the viam office map in localization mode.

Screen Shot 2023-02-09 at 5 57 41 PM

JIRA: https://viam.atlassian.net/browse/RSDK-1785?atlOrigin=eyJpIjoiNjg4NWIwYzRhNTMyNGZiMGI5Y2M2Yzk0YzI0NmIyMzMiLCJwIjoiaiJ9

### DIFF
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -100,11 +100,11 @@ std::atomic<bool> b_continue_session{true};
     if (localization_map_ready) {
         response->set_point_cloud_pcd(localization_pointcloud_map);
         return grpc::Status::OK;
-    }else if(action_mode == ActionMode::LOCALIZING){
-        std::string localization_error = "Localization map is still being prepared";
+    } else if (action_mode == ActionMode::LOCALIZING) {
+        std::string localization_error =
+            "Localization map is still being prepared";
         LOG(ERROR) << localization_error;
-        return grpc::Status(grpc::StatusCode::UNAVAILABLE,
-                                localization_error);
+        return grpc::Status(grpc::StatusCode::UNAVAILABLE, localization_error);
     }
     std::string pointcloud_map;
     // Write or grab the latest pointcloud map in form of a string
@@ -231,11 +231,11 @@ std::atomic<bool> b_continue_session{true};
         common::v1::PointCloudObject *pco = response->mutable_point_cloud();
         pco->set_point_cloud(localization_pointcloud_map);
         return grpc::Status::OK;
-    }else if(action_mode == ActionMode::LOCALIZING){
-        std::string localization_error = "Localization map is still being prepared";
+    } else if (action_mode == ActionMode::LOCALIZING) {
+        std::string localization_error =
+            "Localization map is still being prepared";
         LOG(ERROR) << localization_error;
-        return grpc::Status(grpc::StatusCode::UNAVAILABLE,
-                                localization_error);
+        return grpc::Status(grpc::StatusCode::UNAVAILABLE, localization_error);
     }
 
     std::string pointcloud_map;
@@ -670,8 +670,8 @@ void SLAMServiceImpl::RunSLAM() {
                 latest_map_filename.find(".pbstream")));
 
         // If using the LOCALIZING action mode, cache a copy of the map before
-        // beginning to process data. If cartographer fails to do this, terminate
-        // the program
+        // beginning to process data. If cartographer fails to do this,
+        // terminate the program
         if (action_mode == ActionMode::LOCALIZING) {
             try {
                 GetLatestSampledPointCloudMapString(
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -97,7 +97,7 @@ std::atomic<bool> b_continue_session{true};
     ServerContext *context, const GetPointCloudMapRequest *request,
     GetPointCloudMapResponse *response) {
     // if we have a map for localization, use it
-    if(localization_map_ready){
+    if (localization_map_ready) {
         response->set_point_cloud_pcd(localization_pointcloud_map);
         return grpc::Status::OK;
     }
@@ -219,15 +219,15 @@ std::atomic<bool> b_continue_session{true};
     }
 }
 
-::grpc::Status SLAMServiceImpl::GetCurrentPointCloudMap(const GetMapRequest *request,
-                                                 GetMapResponse *response) {
+::grpc::Status SLAMServiceImpl::GetCurrentPointCloudMap(
+    const GetMapRequest *request, GetMapResponse *response) {
     // if we have a map for localization, use it
-    if(localization_map_ready){
+    if (localization_map_ready) {
         common::v1::PointCloudObject *pco = response->mutable_point_cloud();
         pco->set_point_cloud(localization_pointcloud_map);
         return grpc::Status::OK;
     }
-    
+
     std::string pointcloud_map;
     // Write or grab the latest pointcloud map in form of a string
     try {
@@ -658,18 +658,24 @@ void SLAMServiceImpl::RunSLAM() {
                 latest_map_filename.find(viam::io::filename_prefix) +
                     viam::io::filename_prefix.length(),
                 latest_map_filename.find(".pbstream")));
-        
-        // If using the LOCALIZING action mode, cache a copy of the map before beinging to process data. If cartographer fails to do this, terminate the program
-        if(action_mode == ActionMode::LOCALIZING){
+
+        // If using the LOCALIZING action mode, cache a copy of the map before
+        // beinging to process data. If cartographer fails to do this, terminate
+        // the program
+        if (action_mode == ActionMode::LOCALIZING) {
             try {
-                GetLatestSampledPointCloudMapString(localization_pointcloud_map);
+                GetLatestSampledPointCloudMapString(
+                    localization_pointcloud_map);
             } catch (std::exception &e) {
-                LOG(ERROR) << "Stopping Cartographer: error encoding localized pointcloud map: " << e.what();
+                LOG(ERROR) << "Stopping Cartographer: error encoding localized "
+                              "pointcloud map: "
+                           << e.what();
                 std::terminate();
             }
-            
-            if(localization_pointcloud_map.empty()){
-                LOG(ERROR) << "Stopping Cartographer: error encoding localized pointcloud map: no map points";
+
+            if (localization_pointcloud_map.empty()) {
+                LOG(ERROR) << "Stopping Cartographer: error encoding localized "
+                              "pointcloud map: no map points";
                 std::terminate();
             }
 
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -96,6 +96,7 @@ std::atomic<bool> b_continue_session{true};
 ::grpc::Status SLAMServiceImpl::GetPointCloudMap(
     ServerContext *context, const GetPointCloudMapRequest *request,
     GetPointCloudMapResponse *response) {
+    // if we have a map for localization, use it
     if(localization_map_ready){
         response->set_point_cloud_pcd(localization_pointcloud_map);
         return grpc::Status::OK;
@@ -220,6 +221,7 @@ std::atomic<bool> b_continue_session{true};
 
 ::grpc::Status SLAMServiceImpl::GetCurrentPointCloudMap(const GetMapRequest *request,
                                                  GetMapResponse *response) {
+    // if we have a map for localization, use it
     if(localization_map_ready){
         common::v1::PointCloudObject *pco = response->mutable_point_cloud();
         pco->set_point_cloud(localization_pointcloud_map);
@@ -656,6 +658,8 @@ void SLAMServiceImpl::RunSLAM() {
                 latest_map_filename.find(viam::io::filename_prefix) +
                     viam::io::filename_prefix.length(),
                 latest_map_filename.find(".pbstream")));
+        
+        // If using the LOCALIZING action mode, cache a copy of the map before beinging to process data. If cartographer fails to do this, terminate the program
         if(action_mode == ActionMode::LOCALIZING){
             try {
                 GetLatestSampledPointCloudMapString(localization_pointcloud_map);
--- a/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
+++ b/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
@@ -96,13 +96,12 @@ std::atomic<bool> b_continue_session{true};
 ::grpc::Status SLAMServiceImpl::GetPointCloudMap(
     ServerContext *context, const GetPointCloudMapRequest *request,
     GetPointCloudMapResponse *response) {
-
     std::string pointcloud_map;
     // Write or grab the latest pointcloud map in form of a string
     try {
         std::shared_lock optimization_lock{optimization_shared_mutex,
                                            std::defer_lock};
-        if ((optimization_lock.try_lock())&&(!localization_map_ready)) {
+        if ((optimization_lock.try_lock()) && (!localization_map_ready)) {
             // We are able to lock the optimization_shared_mutex, which means
             // that the optimization is not ongoing and we can grab the newest
             // map
@@ -110,14 +109,17 @@ std::atomic<bool> b_continue_session{true};
             std::lock_guard<std::mutex> lk(viam_response_mutex);
             latest_pointcloud_map = pointcloud_map;
         } else {
-            // Either we are in localization mode or we couldn't lock the mutex which means the optimization process
-            // locked it and we need to use the backed up latest map
-            if(localization_map_ready){
-                LOG(INFO) << "In localization mode, using cached pointcloud map";
-            }else{
-                LOG(INFO) << "Optimization is occuring, using cached pointcloud map";
+            // Either we are in localization mode or we couldn't lock the mutex
+            // which means the optimization process locked it and we need to use
+            // the backed up latest map
+            if (localization_map_ready) {
+                LOG(INFO)
+                    << "In localization mode, using cached pointcloud map";
+            } else {
+                LOG(INFO)
+                    << "Optimization is occuring, using cached pointcloud map";
             }
-            
+
             std::lock_guard<std::mutex> lk(viam_response_mutex);
             pointcloud_map = latest_pointcloud_map;
         }
@@ -221,13 +223,12 @@ std::atomic<bool> b_continue_session{true};
 
 ::grpc::Status SLAMServiceImpl::GetCurrentPointCloudMap(
     const GetMapRequest *request, GetMapResponse *response) {
-
     std::string pointcloud_map;
     // Write or grab the latest pointcloud map in form of a string
     try {
         std::shared_lock optimization_lock{optimization_shared_mutex,
                                            std::defer_lock};
-        if ((optimization_lock.try_lock())&&(!localization_map_ready)) {
+        if ((optimization_lock.try_lock()) && (!localization_map_ready)) {
             // We are able to lock the optimization_shared_mutex, which means
             // that the optimization is not ongoing and we can grab the newest
             // map
@@ -235,12 +236,15 @@ std::atomic<bool> b_continue_session{true};
             std::lock_guard<std::mutex> lk(viam_response_mutex);
             latest_pointcloud_map = pointcloud_map;
         } else {
-            // Either we are in localization mode or we couldn't lock the mutex which means the optimization process
-            // locked it and we need to use the backed up latest map
-            if(localization_map_ready){
-                LOG(INFO) << "In localization mode, using cached pointcloud map";
-            }else{
-                LOG(INFO) << "Optimization is occuring, using cached pointcloud map";
+            // Either we are in localization mode or we couldn't lock the mutex
+            // which means the optimization process locked it and we need to use
+            // the backed up latest map
+            if (localization_map_ready) {
+                LOG(INFO)
+                    << "In localization mode, using cached pointcloud map";
+            } else {
+                LOG(INFO)
+                    << "Optimization is occuring, using cached pointcloud map";
             }
 
             std::lock_guard<std::mutex> lk(viam_response_mutex);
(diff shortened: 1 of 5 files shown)


slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -646,6 +658,29 @@ void SLAMServiceImpl::RunSLAM() {
                 latest_map_filename.find(viam::io::filename_prefix) +
                     viam::io::filename_prefix.length(),
                 latest_map_filename.find(".pbstream")));
+
+        // If using the LOCALIZING action mode, cache a copy of the map before
+        // beinging to process data. If cartographer fails to do this, terminate
```
tessavitabile, line 663: "beinging" should be "beginning"

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -292,6 +292,12 @@ class SLAMServiceImpl final : public SLAMService::Service {
     std::string latest_jpeg_map_without_marker;
 
     std::string latest_pointcloud_map;
+
+    // Enable Localization mode to utilize a cached version of the pointcloud
+    // map, to skip unecessary encoding steps
+    std::atomic<bool> localization_map_ready{false};
+    std::string localization_pointcloud_map;
```
tessavitabile, line 299: nit: these should be declared above the mutexes, so it is clear they are not guarded by the mutexes.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -96,6 +96,11 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
 ::grpc::Status SLAMServiceImpl::GetPointCloudMap(
     ServerContext *context, const GetPointCloudMapRequest *request,
     GetPointCloudMapResponse *response) {
+    // if we have a map for localization, use it
+    if (localization_map_ready) {
```
tessavitabile, line 100: If we're in localization mode but the localization map is not ready, we should return UNAVAILABLE here. Otherwise, we might paint a map (which would contain the new submaps used for localization).

tessavitabile (approved):
Nice work! A few comments.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
JohnN193: will do, figured since I was treating the atomic bool as a lock that it should live down there

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: kk added a check for if `localization_map_ready == false` and we are in localization mode

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -96,6 +96,16 @@ ::grpc::Status SLAMServiceImpl::GetPositionNew(
 ::grpc::Status SLAMServiceImpl::GetPointCloudMap(
     ServerContext *context, const GetPointCloudMapRequest *request,
     GetPointCloudMapResponse *response) {
+    // if we have a map for localization, use it
```
kkufieta, line 99: [nit] for consistency with rest of comments

```suggestion
    // If we have a map for localization, use it
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -216,6 +226,18 @@ ::grpc::Status SLAMServiceImpl::GetJpegMap(const GetMapRequest *request,
 
 ::grpc::Status SLAMServiceImpl::GetCurrentPointCloudMap(
     const GetMapRequest *request, GetMapResponse *response) {
+    // if we have a map for localization, use it
```
kkufieta, line 229: [nit] for consistency with rest of comments

```suggestion
    // If we have a map for localization, use it
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -269,6 +269,11 @@ class SLAMServiceImpl final : public SLAMService::Service {
     size_t current_file_offline = 0;
     std::string current_file_online;
 
+    // Enable Localization mode to utilize a cached version of the pointcloud
+    // map, to skip unecessary encoding steps
+    std::atomic<bool> localization_map_ready{false};
+    std::string localization_pointcloud_map;
```
kkufieta, line 275: [q] Why not use `latest_pointcloud_map` instead of creating a new variable?

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -646,6 +668,29 @@ void SLAMServiceImpl::RunSLAM() {
...
+            }
+
+            if (localization_pointcloud_map.empty()) {
+                LOG(ERROR) << "Stopping Cartographer: error encoding localized "
+                              "pointcloud map: no map points";
+                std::terminate();
+            }
+
+            localization_map_ready = true;
+        }
```
kkufieta, line 693: [q] Could we run this for both updating & localizing mode and save the map in `latest_pointcloud_map` instead? That way, we'd also have a map for "updating" at startup - and when no data has been added yet (e.g., maybe a user wants to view the map they load without running cartographer with new data?)

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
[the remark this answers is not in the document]
JohnN193: the behavior of the `GetPosition` endpoint is not related to this ticket. If we had a way to determine when localization occurred we could maybe add some debug logs to inform users that GetPosition might not be accurate. 

I see no reason to modify the position endpoint, as Cartographer always assumes the robot will start from the origin when in localizing/updating mode, unless localization actually occurs

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -216,6 +226,18 @@ ::grpc::Status SLAMServiceImpl::GetJpegMap(const GetMapRequest *request,
 
 ::grpc::Status SLAMServiceImpl::GetCurrentPointCloudMap(
     const GetMapRequest *request, GetMapResponse *response) {
+    // if we have a map for localization, use it
+    if (localization_map_ready) {
```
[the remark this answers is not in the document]
JohnN193: I did not turn this into a helper function because the `GetMap` endpoint should be removed soon, but I can go ahead and make a function based off your other feedback

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
JohnN193: I made a new variable because it's not the `latest_pointcloud_map`. I can rewrite some logic to use this instead

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: There's no reason to run at startup in updating mode. They could just ask for a map and receive the same result. In addition, there is no way for updating mode to know whether to use the backup map vs just asking for a new map and to make it smart enough for the scenario you suggest is out of scope for this ticket.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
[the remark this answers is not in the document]
JohnN193: I remembered the real reason, `GetMap` and `GetPointCloudMap` have two different response types, so I would need to make a different function for each one.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -101,18 +101,25 @@ ::grpc::Status SLAMServiceImpl::GetPointCloudMap(
     try {
         std::shared_lock optimization_lock{optimization_shared_mutex,
                                            std::defer_lock};
-        if (optimization_lock.try_lock()) {
+        if ((optimization_lock.try_lock()) && (!localization_map_ready)) {
```
tessavitabile, line 104: [nit] remove extra parentheses

tessavitabile, line 104: You should check `localization_map_ready` before trying to acquire the lock (i.e. put it first in the if-statement).

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -221,18 +228,25 @@ ::grpc::Status SLAMServiceImpl::GetCurrentPointCloudMap(
...
```
tessavitabile, line 231: Same here, switch the order of checking localization_map_ready and locking, and remove extra parens.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -888,6 +935,8 @@ void SLAMServiceImpl::ProcessDataAndStartSavingMaps(double data_start_time) {
             std::lock_guard<std::mutex> lk(viam_response_mutex);
             latest_global_pose = tmp_global_pose;
         }
+        if (action_mode == ActionMode::LOCALIZING) BackupLatestMap();
```
tessavitabile, line 938: I don't understand we we need to call BackupLatestMap here, since the purpose of BackupLatestMap is to save a cached map to serve during optimization, and at this point, we're done with optimization.

This also makes concurrency unsafe, since BackupLatestMap will modify latest_pointcloud_map, but now it's possible to access latest_pointcloud_map without holding optimization_lock, as long as localization_map_ready is true.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -866,7 +913,7 @@ void SLAMServiceImpl::ProcessDataAndStartSavingMaps(double data_start_time) {
         map_builder.map_builder_->FinishTrajectory(trajectory_id);
     }
     if (!use_live_data) {
-        BackupLatestMap();
+        if (action_mode != ActionMode::LOCALIZING) BackupLatestMap();
         {
             std::unique_lock<std::shared_mutex> optimization_lock(
```
tessavitabile, line 918: Do we need to optimize the map at all in localization mode?

tessavitabile (changes requested):
I think this change makes the concurrency too hard to reason about. Now the rule is that you can access latest_pointcloud_map if you are holding optimization_lock OR localization_map_ready is true, which is just too complicated to follow correctly. Having separate variables for latest_pointcloud_map and the localization_map was clearer and had simpler concurrency logic, so I think we should go back to that.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: `BackupLatestMap` holds the `viam_response_mutex`, so I figured that this would still be safe to do. Thinking about this again I probably do not need to run this for localization mode at all.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: I would say yes because the optimization step can help get cartographer to localize on the map. In the map/dataset I have posted above, the jpeg map/pose fixes itself(i.e. looks aligned) after the optimization step occurs.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: Never mind, I missed that the viam_response_mutex is responsible for controlling access to latest_pointcloud_map, so I think this is safe. I still suggest not calling BackupLatestMap here, if we don't need it.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: thanks, can you add that as a comment?

tessavitabile (approved):

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -101,18 +101,25 @@ ::grpc::Status SLAMServiceImpl::GetPointCloudMap(
...
             latest_pointcloud_map = pointcloud_map;
         } else {
-            // We couldn't lock the mutex which means the optimization process
-            // locked it and we need to use the backed up latest map
-            LOG(INFO)
-                << "Optimization is occuring, using cached pointcloud map";
+            // Either we are in localization mode or we couldn't lock the mutex
+            // which means the optimization process locked it and we need to use
+            // the backed up latest map
+            if (localization_map_ready) {
```
kkufieta, line 115: [minor]
```suggestion
            if (action_mode == ActionMode::LOCALIZING) {
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta, line 231: [minor]
```suggestion
        if (action_mode != ActionMode::LOCALIZING && optimization_lock.try_lock()) {
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -221,18 +228,25 @@ ::grpc::Status SLAMServiceImpl::GetCurrentPointCloudMap(
...
```
kkufieta, line 242: [minor]
```suggestion
            if (action_mode == ActionMode::LOCALIZING) {
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -340,6 +354,37 @@ void SLAMServiceImpl::BackupLatestMap() {
     latest_pointcloud_map = std::move(pointcloud_map_tmp);
 }
 
+// If using the LOCALIZING action mode, cache a copy of the map before
+// beginning to process data. If cartographer fails to do this,
+// terminate the program
+bool SLAMServiceImpl::CacheMapInLocalizationMode() {
```
kkufieta, line 360: [minor]
```suggestion
void SLAMServiceImpl::CacheMapInLocalizationMode() {
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -340,6 +354,37 @@ void SLAMServiceImpl::BackupLatestMap() {
...
+            std::terminate();
+        }
+
+        {
+            std::lock_guard<std::mutex> lk(viam_response_mutex);
+            latest_pointcloud_map = std::move(pointcloud_map_tmp);
+        }
+        return true;
+    }
+    return false;
```
kkufieta, line 385: [minor]
```suggestion
    }
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -646,6 +691,8 @@ void SLAMServiceImpl::RunSLAM() {
                 latest_map_filename.find(viam::io::filename_prefix) +
                     viam::io::filename_prefix.length(),
                 latest_map_filename.find(".pbstream")));
+
+        localization_map_ready = CacheMapInLocalizationMode();
```
kkufieta, line 695: [minor]
```suggestion
        CacheMapInLocalizationMode();
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: Good question, I wonder as well if we should have this instead:

```cpp
    if (action_mode != ActionMode::LOCALIZING && !use_live_data) {
        BackupLatestMap();
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -888,6 +935,8 @@ void SLAMServiceImpl::ProcessDataAndStartSavingMaps(double data_start_time) {
             std::lock_guard<std::mutex> lk(viam_response_mutex);
             latest_global_pose = tmp_global_pose;
         }
+        if (action_mode == ActionMode::LOCALIZING) BackupLatestMap();
+
```
kkufieta, line 939: [minor]
```suggestion
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -258,6 +258,11 @@ class SLAMServiceImpl final : public SLAMService::Service {
     // the respective member variables.
     void BackupLatestMap();
 
+    // If using the LOCALIZING action mode, cache a copy of the map before
+    // beginning to process data. If cartographer fails to do this,
+    // terminate the program
+    bool CacheMapInLocalizationMode();
```
kkufieta, line 264: [minor]

```suggestion
    // terminate the program.
    void CacheMapInLocalizationMode();
```

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -269,6 +274,10 @@ class SLAMServiceImpl final : public SLAMService::Service {
     size_t current_file_offline = 0;
     std::string current_file_online;
 
+    // Enable Localization mode to utilize a cached version of the pointcloud
+    // map, to skip unecessary encoding steps
+    std::atomic<bool> localization_map_ready{false};
+
```
kkufieta, line 280: [minor]

```suggestion
```

kkufieta (changes requested):
As we've discussed offline, I left a handful of [minor] comments to remove `localization_map_ready` and rely directly on checking whether or not `action_mode == ActionMode::LOCALIZING`. The comments are [minor] as long as there aren't good reasons why we should keep `localization_map_ready`.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: EDIT: Just saw the discussion that happened 40 min ago, so my comment is a outdated.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
kkufieta: This is [opt] now, given the discussion that happened 40 min ago.

JohnN193:
> As we've discussed offline, I left a handful of [minor] comments to remove `localization_map_ready` and rely directly on checking whether or not `action_mode == ActionMode::LOCALIZING`. The comments are [minor] as long as there aren't good reasons why we should keep `localization_map_ready`.

I picked an atomic bool because I figured it would be safer, so we did not need a mutex lock `action_mode`. Changing this now as it seems that is not necessary

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
nicksanford: I agree with Kat's suggestion. To me, whether cartogropher is in localization mode should only be determined  exclusively by the action mode.

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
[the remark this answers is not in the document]
JohnN193: was already making changes so I added this to my commit

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: was already making changes so I added this to my commit

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: was already making changes so I added this to my commit

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: was already making changes so I added this to my commit

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: was already making changes so I added this to my commit

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: was already making changes so I added this to my commit

slam-libraries/viam-cartographer/src/slam_service/slam_service.cc
JohnN193: was already making changes so I added this to my commit

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -251,13 +251,18 @@ class SLAMServiceImpl final : public SLAMService::Service {
...
     // limit on gRPC messages. The sampled behavior may change when moving to
-    // streamed point clouds
+    // streamed point clouds.
     void GetLatestSampledPointCloudMapString(std::string &pointcloud);
 
     // BackupLatestMap extracts and saves the latest map as a backup in
     // the respective member variables.
     void BackupLatestMap();
 
+    // If using the LOCALIZING action mode, cache a copy of the map before
```
nicksanford, line 261: [nit] can we either have the comment for this function be on the prototype or the function definition but not both? Otherwise, it makes it likely that the two comments will diverge over time, leading to confusion.

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
JohnN193: was already making changes so I added this to my commit

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
JohnN193: was already making changes so I added this to my commit

slam-libraries/viam-cartographer/src/slam_service/slam_service.h
JohnN193: meant to only have it on the prototype, removed it from the definition

nicksanford (approved):
Thanks for the refactors, this looks good to me (mod comment nit). Very clear & easy to follow.

Also thanks @kkufieta for the great suggestions.

kkufieta (approved):
LGTM - Thanks for accepting the suggested changes!

PR #149, opened 2023-02-17
[RSDK-1955] add orbslam GetInternalStateStream

JohnN193:
ticket: [RSDK-1955 add orbslam GetInternalStateStream](https://viam.atlassian.net/browse/RSDK-1955?atlOrigin=eyJpIjoiODcyMzc2MzBlNDY2NDk4NWI5MjdjNWY2YjYwMGU5MDQiLCJwIjoiaiJ9)

adds the GetInternalStateStream endpoint to for orbslam. 

testing:
run orbslam using a euroc mono camera dataset. 

first running orb-grpc-server and pinging the endpoint, and comparing to an `.osa` file saved with `map_rate_sec`. We found that on multiple save map calls orbslam will add 6-12 bytes per call. Unary and stream endpoints were pinged multiple times

```
johnn193@john5-pi:~/api $ ../go/bin/grpcurl -max-msg-sz 10000000 -d '{"name": "slam"}' -plaintext -protoset <(~/slam/slam-libraries/grpc/bin/buf build -o -) $(sudo netstat -tulpn | grep LISTEN | grep orb_grpc_serv | awk '{print $4}' | sort | head -n 1) viam.service.slam.v1.SLAMService/GetInternalStateStream | jq -r ".internalStateChunk"  | ./base64_decode_lines.py > testStream.osa

johnn193@john5-pi:~/api $ ls -l testStream.osa 
-rw-r--r-- 1 johnn193 johnn193 37971684 Feb 21 17:45 testStream.osa

johnn193@john5-pi:~/api $ ls -l ../euroc_MH_01/map/color_data_2023-02-21T17\:37\:20.0000Z.osa 
-rw-r--r-- 1 root root 37971672 Feb 21 17:37 ../euroc_MH_01/map/color_data_2023-02-21T17:37:20.0000Z.osa
```

then I made sure orbslam could successfully load the .osa files. 
```
2023-02-21T18:27:14.263Z	INFO	robot_server.process.slam_orbslamv3_orb_grpc_server.StdOut	pexec/managed_process.go:221	
\_ Load File
2023-02-21T18:27:14.265Z	INFO	robot_server.process.slam_orbslamv3_orb_grpc_server.StdOut	pexec/managed_process.go:221	
\_ Initialization of Atlas from file: /home/johnn193/api/testStream
2023-02-21T18:27:14.267Z	INFO	robot_server.process.slam_orbslamv3_orb_grpc_server.StdOut	pexec/managed_process.go:221	
\_ Starting to read the save binary file
2023-02-21T18:27:15.067Z	INFO	robot_server.process.slam_orbslamv3_orb_grpc_server.StdOut	pexec/managed_process.go:221	

\_ End to load the save binary file
```


slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
```diff
@@ -439,6 +439,38 @@ ::grpc::Status SLAMServiceImpl::GetInternalState(
...
+    // deferring reading the osa file in chunks until we run into issues
+    // with loading the file into memory
+    bool success = ArchiveSlam(buffer);
+    if (!success)
+        return grpc::Status(grpc::StatusCode::UNAVAILABLE,
+                            "SLAM is not yet initialized");
+
+    std::string internal_state_chunk;
+    GetInternalStateStreamResponse response;
+    for (int start_index = 0; start_index < buffer.str().size();
```
tessavitabile, line 461: We should avoid calling `buffer.str().size()` on every iteration of the loop, since I believe that will copy the buffer to a string on each iteration. Instead, we can call this once before the loop starts, and save it as a const.

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -30,6 +35,10 @@ namespace viam {
 static const int filenamePrefixLength = 6;
 static const int checkForShutdownIntervalMicroseconds = 1e5;
 extern std::atomic<bool> b_continue_session;
+// Byte limit on GRPC, used to help determine sampling skip_count
```
tessavitabile, line 38: This comment doesn't apply to orbslam.

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -30,6 +35,10 @@ namespace viam {
...
+static const int maximumGRPCByteLimit = 32 * 1024 * 1024;
+// Byte limit for chunks on GRPC, used for streaming apis
+static const int maximumGRPCByteChunkSize = 64 * 1024;
```
tessavitabile, line 41: Why do we use a smaller value for maximumGRPCByteChunkSize than maximumGRPCByteLimit?

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -64,6 +73,20 @@ class SLAMServiceImpl final : public SLAMService::Service {
         ServerContext *context, const GetInternalStateRequest *request,
         GetInternalStateResponse *response) override;
 
+    // GetPointCloudMap returns a stream of the current sampled pointcloud
+    // derived from the painted map, using probability estimates in chunks with
```
tessavitabile, line 77: This comment isn't correct for orbslam

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -64,6 +73,20 @@ class SLAMServiceImpl final : public SLAMService::Service {
...
+    // a max size of maximumGRPCByteChunkSize
+    ::grpc::Status GetPointCloudMapStream(
+        ServerContext *context, const GetPointCloudMapStreamRequest *request,
+        ServerWriter<GetPointCloudMapStreamResponse> *writer) override;
+
+    // GetInternalState returns a stream of the current internal state of the
+    // map which is a pbstream for cartographer in chunks of size
```
tessavitabile, line 84: This comment isn't correct for orbslam

tessavitabile (changes requested):

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
nicksanford: The maximumGRPCByteLimit is 32 MB which is quite large.  I recommended smaller ByteChunkSizes of 64 kb so that we minimize memory consumption of the buffered response both on the slam algo side & the RDK caller side.

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
nicksanford: @tessavitabile 
Does that mean we should also not call `buffer.str().substr` here as well https://github.com/viamrobotics/slam/pull/149/files#diff-b68d96da1bc970406f0f4402811d582812550e7c8450a636d73540f0363c73d8R464 ?

Should we should instead do something like:

```
string buffer_str = buffer.str()
for (int start_index = 0; start_index < buffer_str.size(); start_index += maximumGRPCByteChunkSize) {
internal_state_chunk = buffer_str.substr(start_index, maximumGRPCByteChunkSize);
...
}
```

tessavitabile: yes

JohnN193: c++ question, can I use `std::move(buffer.str())` instead of making a copy? since `buffer` would not get used in the rest of the code

nicksanford (approved):

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
```diff
@@ -50,20 +59,34 @@ class SLAMServiceImpl final : public SLAMService::Service {
...
     ::grpc::Status GetPointCloudMap(
         ServerContext *context, const GetPointCloudMapRequest *request,
         GetPointCloudMapResponse *response) override;
 
     // For a given GetInternalStateRequest
     // Returns a GetInternalStateResponse containing
-    // the internal state of the SLAM algorithm
-    // required to continue mapping/localization
+    // current internal state of the map represented as an ORB-SLAM Atlas(.osa)
+    // file in chunks of size maximumGRPCByteChunkSize
```
tessavitabile, line 71: This doesn't return chunks, but it's not so important, since this is going away soon

tessavitabile (approved):

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
nicksanford: I don't really know much about `std::move` but from reading the docs: https://en.cppreference.com/w/cpp/utility/move I do think that it should be safe to change `std::string buffer_str = buffer.str();` to `std::string buffer_str = std::move(buffer.str());` given you don't use it again.

nicksanford: If you do change it, please run the manual tests to confirm that it hasn't changed the behavior.

jeremyhyde-viam: I think since the `string buffer_str = buffer.str()` includes the conversion to string, move is not actually duplicating the memory allocation. The conversion creates a string object that is then assigned to buffer_str. `std::move()` seems to be used more when copying variables of the same type and just want to send the pointer instead of the full object.

jeremyhyde-viam (approved):
LGTM, left some thoughts on the `std::move` question, but otherwise logic is good!

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
jeremyhyde-viam: We increased from 4MB to 32MB to allow pointclouds/maps beyond 4MB to be sent. With streaming that reason no longer holds.

However general consensus seems to be: "The 4 MB limit is protect clients/servers who haven't thought about message size constraints. gRPC itself is fine with going much higher (100s of MBs), but most applications could be trivially attacked or accidentally go out-of-memory allowing messages of that size.

If you're willing to receive a 100 MB message all-at-once, then increasing the limit is fine."

We should be able to change the chunk size to 32MB if we are confident that no 3rd party apps, that will need to ingest the streaming data, have data constraints on them.

nicksanford: re `If we are using a smaller chunk size, then why do we increase the max message size from 4MB to 32MB? Is this just to be consistent with rdk?` 

I don't see where we are changing the max unary message size from 4MB to 32BM. Where do you see that?

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
tessavitabile: Yes, std::move isn't necessary here, since buffer.str() is already an rvalue (unnamed value), so it will not be copied when it is assigned to buffer_str. std::move is when you want to move ownership from one lvalue (named value) to another. This doesn't share a pointer or create shared ownership--instead the moved-from value will be unsafe to access.

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
tessavitabile: The changes in [slam-libraries/viam-orb-slam3/orbslam_server_v1_main.cc](https://github.com/viamrobotics/slam/pull/149/files#diff-3d6d11e825448c5db341a7bc8b68686e579dcf74c77b90b9638fcc05358f5e75)

Can we remove the line to increase the max message size after the streaming changes are done?

slam-libraries/viam-orb-slam3/orbslam_server_v1.cc
JohnN193: kk won't make the change then, thanks!

slam-libraries/viam-orb-slam3/orbslam_server_v1.h
nicksanford: Sorry I might be missing something. I think John is just moving a magic number into a constant with the same value.

The change is going from:
```
   builder.SetMaxSendMessageSize(32 * 1024 * 1024);
```

to
```
builder.SetMaxSendMessageSize(viam::maximumGRPCByteLimit);
```

where `viam::maximumGRPCByteLimit` is equal to 
```
static const int maximumGRPCByteLimit = 32 * 1024 * 1024;
```

So I don't think this is going to change any behavior at all. Am I still missing something?

tessavitabile: No, it's not a behavior change. It was just something I noticed as part of the code review.

JohnN193: oh yeah, I realized we had a magic number in that spot so I swapped to using the variable that was consistent with the cartographer implementation

nicksanford: Got it. Understood. Sorry for my confusion.

PR #151, opened 2023-02-21
[RSDK-2002] Copy/paste internal/config & related tests from rdk into slam repo

kkufieta:
https://viam.atlassian.net/browse/RSDK-2002

* Copy/paste internal/config from rdk into slam repo
* Copy/paste related tests from builtin_test.go from rdk into slam repo


tessavitabile (approved):
The ticket also includes adding CI for config_test.go. Will that be in a follow up PR?

kkufieta:
@tessavitabile yes, I was thinking of doing that in a follow-up ticket. Merging this ticket is going to make that easier, as well as allow merging necessary changes in RDK.

go.sum
```diff
@@ -0,0 +1,981 @@
...
+cloud.google.com/go v0.45.1/go.mod h1:RpBamKRgapWJb87xiFSdk4g1CME7QZg3uwTez+TSTjc=
+cloud.google.com/go v0.46.3/go.mod h1:a6bKKbmY7er1mI7TEI4lsAkts/mkhTSZK8w33B4RAg0=
+cloud.google.com/go v0.50.0/go.mod h1:r9sluTvynVuxRIOHXQEHMFffphuXHOMZMycpNR5e6To=
+cloud.google.com/go v0.52.0/go.mod h1:pXajvRH/6o3+F9jDHZWQ5PbGhn+o8w9qiu/CffaVdO4=
+cloud.google.com/go v0.53.0/go.mod h1:fp/UouUEsRkN6ryDKNW/Upv/JBKnv6WDthjR6+vze6M=
+cloud.google.com/go v0.54.0/go.mod h1:1rq2OEkV3YMf6n/9ZvGWI3GWw0VoqH/1x2nd8Is/bPc=
+cloud.google.com/go v0.56.0/go.mod h1:jr7tqZxxKOVYizybht9+26Z/gUq7tiRzu+ACVAMbKVk=
+cloud.google.com/go v0.57.0/go.mod h1:oXiQ6Rzq3RAkkY7N6t3TcE6jE+CIBBbA36lwQ1JyzZs=
+cloud.google.com/go v0.60.0/go.mod h1:yw2G51M9IfRboUH61Us8GqCeF1PzPblB823Mn2q2eAU=
+cloud.google.com/go/bigquery v1.0.1/go.mod h1:i/xbL2UlR5RvWAURpBYZTtm/cXjCha9lbfbpx4poX+o=
```
nicksanford, line 17: 1. Are all the package versions in go.sum the same as they were in RDK?
2. Why do we need to install things like `cloud.google.com/go/bigquer` for this simple program?

nicksanford (approved):

nicksanford:
I'm approving but would like to either trim unnecessary dependencies if we have any & keep the same dependency versions we have in RDK during the modularization migration.

go.sum
kkufieta: I am not a go expert, but as far as I understand it, `go mod tidy` pulls & cleans up dependencies to be the minimally required ones for the imported packages. `go.sum` is autogenerated after running `go mod tidy`. This means my answers to your questions are:

1. I don't believe the go.sum is the same as in RDK
2. I don't know why we need to install `bigquery`, but I assume one of the packages we're importing relies on it under the hood.

go.mod
```diff
@@ -0,0 +1,17 @@
+module go.viam.com/slam
+
+go 1.19
+
+require (
+	github.com/edaniels/golog v0.0.0-20230215213219-28954395e8d0
+	github.com/pkg/errors v0.9.1
+	go.viam.com/test v1.1.0
```
nicksanford, line 8: Is there a reason why we are using a different version than rdk? https://github.com/viamrobotics/rdk/blob/main/go.mod#L86

go.mod
```diff
@@ -0,0 +1,17 @@
...
+)
+
+require (
+	github.com/benbjohnson/clock v1.1.0 // indirect
```
nicksanford, line 12: Is there a reason why we are using a different version than RDK? https://github.com/viamrobotics/rdk/blob/main/go.mod#L141d

go.mod
```diff
@@ -0,0 +1,17 @@
...
+	github.com/smartystreets/assertions v1.2.0 // indirect
+	go.uber.org/atomic v1.7.0 // indirect
```
nicksanford, line 14: Is there a reason why we are using a different version than RDK?
https://github.com/viamrobotics/rdk/blob/main/go.mod#L81

go.mod
```diff
@@ -0,0 +1,17 @@
...
+	go.uber.org/multierr v1.6.0 // indirect
```
nicksanford, line 15: Is there a reason why we are using a different version than RDK?
https://github.com/viamrobotics/rdk/blob/main/go.mod#L82

go.mod
```diff
@@ -0,0 +1,17 @@
...
+	go.uber.org/zap v1.23.0 // indirect
```
nicksanford, line 16: Is there a reason why we are using a different version than RDK?
https://github.com/viamrobotics/rdk/blob/main/go.mod#L83

nicksanford (changes requested):
Got some questions about the versions of packages we are installing & why they differ from RDK

nicksanford:
> I'm approving but would like to either trim unnecessary dependencies if we have any & keep the same dependency versions we have in RDK during the modularization migration.

Actually on further investigation I found that we are indeed using different dependency versions than RDK & I recommend pinning the versions to the same as RDK before merging.

zaporter-work (approved):
LGTM (With Nick's dependency concerns)

go.mod
edaniels: could have it match. it probably doesn't matter. I'm assuming all things here took latest.

edaniels:
I don't think matching the dependencies matters much unless it causes some issue. You can just check what you're missing between the two versions which is ShouldResembleProto and ShouldResemble handling some proto changes. If you don't need them, being strict about matching versions is an effortful process. If we really care about that across repositories, we need a syncing solution.

nicksanford (approved):
Approved following eric's comment.

PR #152, opened 2023-02-21
[RSDK-2002] Add testing workflow automation

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-2002

* Set up github testing workflow & actions
* Fix a small bug from previous commit (make newError private)


.github/workflows/main.yml
```diff
@@ -17,9 +17,12 @@ on:
...
-    secrets:
-      GCP_CREDENTIALS: ${{ secrets.GCP_CREDENTIALS }}
-      REPO_READ_TOKEN: ${{ secrets.REPO_READ_TOKEN }}
-      ARTIFACT_READ_ONLY_GCP_CREDENTIALS: ${{ secrets.ARTIFACT_READ_ONLY_GCP_CREDENTIALS }}
+  # appimage:
+  #   uses: viamrobotics/slam/.github/workflows/appimage.yml@main
+  #   secrets:
+  #     GCP_CREDENTIALS: ${{ secrets.GCP_CREDENTIALS }}
+  #     REPO_READ_TOKEN: ${{ secrets.REPO_READ_TOKEN }}
+  #     ARTIFACT_READ_ONLY_GCP_CREDENTIALS: ${{ secrets.ARTIFACT_READ_ONLY_GCP_CREDENTIALS }}
```
kkufieta, line 25: Will uncomment again once this PR is approved

.github/workflows/main.yml
```diff
@@ -6,7 +6,7 @@ concurrency:
 on:
   workflow_dispatch:
   push:
-    branches: [ 'main' ]
+    branches: [ 'kk/add-lint-test' ]
```
kkufieta, line 9: All mentions of `kk/add-lint-test` will be renamed to `main` once the PR is approved.

.github/workflows/appimage.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       options: --platform ${{ matrix.platform }}
     timeout-minutes: 180
     outputs:
-      date: ${{ steps.build_test_app.outputs.date }}
+      date: ${{ steps.build_appimage.outputs.date }}
```
Otterverse, line 40: Where is this output being used? I can't seem to find it, but maybe I'm missing something.

.github/workflows/appimage.yml
```diff
@@ -103,7 +103,10 @@ jobs:
         sudo --preserve-env=GOOGLE_APPLICATION_CREDENTIALS -u testbot bash -lc 'cd rdk/services/slam/builtin && sudo go test -v -run TestOrbslamIntegration'
 
     - name: Build appimage
-      run: sudo -u testbot bash -lc 'cd slam/slam-libraries && make BUILD_CHANNEL="latest" appimageorb'
+      id: build_appimage
+      run: |
+        echo "date=`date +%F`" >> $GITHUB_OUTPUT
```
Otterverse, line 108: same question, where is the output actually being used?

Otterverse: Ditto for two other places where the date output is being set.

Otterverse (commented):
Looks good for the most part, just one outstanding question about where the "date" output is being used. If it's not used, would like to see it removed.

.github/workflows/appimage.yml
kkufieta: I don't have an answer for this, to be honest. We changed this because we added the action linter and had to change `build_test_app` to `build_appimage` in order for the linter to be happy. While we were at it, we added `echo "date=`date +%F`" >> $GITHUB_OUTPUT` because that's what is being done in RDK, but we did wonder if we needed `date` altogether.

Given that you think it's unused and should be removed, I will do that.

Otterverse: Yeah, if there's not a job elsewhere that depends on this job (AND actually uses that output) then it can go away and simplify the file.

In RDK, it's because a later job actually tests the built appimage, and needs the date to get the right URL for the temporary file (since computer clocks might not be in the same timezone, or tests might happen near midnight.)

Otterverse (approved):
LGTM! Glad to see a linter getting put in place, at least for the go code for now.

.github/workflows/appimage.yml
```diff
@@ -36,8 +36,6 @@ jobs:
       image: ${{ matrix.image }}
       options: --platform ${{ matrix.platform }}
     timeout-minutes: 180
-    outputs:
-      date: ${{ steps.build_test_app.outputs.date }}
```
tessavitabile, line 40: What is the reason for this change?

.github/workflows/gotest.yml
```diff
@@ -0,0 +1,73 @@
...
+      matrix:
+        include:
+          - arch: [x64, qemu-host]
+            image: ghcr.io/viamrobotics/canon:amd64-cache
+            platform: linux/amd64
+            lint: 'make lint'
+          - arch: [arm64, qemu-host]
+            image: ghcr.io/viamrobotics/canon:arm64-cache
+            platform: linux/arm64
+            lint: 'make lint'
```
tessavitabile, line 20: [nit] I don't think we need to define 'lint' here, since it's the same for both platforms.

tessavitabile (approved):
LGTM with a question and a nit

.github/workflows/appimage.yml
kkufieta: The date is unused

.github/workflows/gotest.yml
kkufieta: Does this mean I would remove this line and line 16, and do the following?

Change this:
```yml
    - name: Verify no uncommitted changes from "${{ matrix.lint }}"
      run: |
        git init
        git add .
        chown -R testbot .
        sudo -u testbot bash -lc '${{ matrix.lint }}'
        GEN_DIFF=$(git status -s)
```

to this?

```yml
    - name: Verify no uncommitted changes from "make lint"
      run: |
        git init
        git add .
        chown -R testbot .
        sudo -u testbot bash -lc make lint
        GEN_DIFF=$(git status -s)
```

JohnN193 (approved):
LGTM

.github/workflows/gotest.yml
tessavitabile: yes

zaporter-work (approved):
LGTM

.github/workflows/gotest.yml
kkufieta: @Otterverse Do I have to put `make lint` in between ticks?

Referring to that, the above works. Specifically, this part: 
```yml
sudo -u testbot bash -lc make lint
```

Whereas for the tests, we have this:
```yml
    - name: Test
      run: |
        sudo --preserve-env=MONGODB_TEST_OUTPUT_URI,GITHUB_SHA,GITHUB_RUN_ID,GITHUB_RUN_NUMBER,GITHUB_RUN_ATTEMPT,GITHUB_X_PR_BASE_SHA,GITHUB_X_PR_BASE_REF,GITHUB_X_HEAD_REF,GITHUB_X_HEAD_SHA,GITHUB_REPOSITORY -u testbot bash -lc 'make test'
```

where we have `'make test'` instead of `make test`. Why and when would I put `make lint` and `make test` between ticks?

nicksanford (approved):

.github/workflows/gotest.yml
Otterverse: So generally, you need to quote the string passed to bash with the -c option. You can test this yourself in the command line.

`bash -c echo fotbar` returns nothing.
`bash -c "echo foobar"` or `bash -c 'echo foobar'` should both actually print out "foobar"

(The discussion over WHICH quote marks to use is another topic, but for simple things with no special characters, either works.)

If make lint was working without the quotes, then it's likely JUST running "make" and not actually passing the second word. That might still work in your case, if the default "make" thing does the same stuff (or does nothing, and thus nothing changes and it still passes the test.)

kkufieta: Great, good to know! Glad I asked :) Will fix now.

PR #159, opened 2023-02-22
[RSDK-2003] Move & update AttrConfig

zaporter-work:
https://viam.atlassian.net/browse/RSDK-2003

Move AttrConfig code & tests from builtin to slam repo


zaporter-work:
Note: This makes the following fields required: use_live_data, data_dir, and delete_processed_data. 
I am not sure if delete_processed_data should be required, and if it is not, I will fix it tomorrow. I only made it required because DetermineDeleteProcessed data causes a segfault if it is not.

kkufieta:
> Note: This makes the following fields required: use_live_data, data_dir, and delete_processed_data.
> I am not sure if delete_processed_data should be required, and if it is not, I will fix it tomorrow. I only made it required because DetermineDeleteProcessed data causes a segfault if it is not.

The following should be true:
* `use_live_data` is required
* `data_dir` is required
* `delete_processed_data` is optional

config/config.go
```diff
@@ -39,3 +41,96 @@ func DetermineUseLiveData(logger golog.Logger, liveData *bool, sensors []string)
...
+		config.Port = port
+	}
+
+	if config.DataRateMs == 0 {
+		config.DataRateMs = defaultDataRateMs
+		logger.Debugf("no data_rate_msec given, setting to default value of %d", defaultDataRateMs)
+	}
+
+	if config.MapRateSec == nil {
+		logger.Debugf("no map_rate_secs given, setting to default value of %d", defaultMapRateSec)
```
tessavitabile, line 121: map_rate_secs should be map_rate_sec

config/config.go
```diff
@@ -39,3 +41,96 @@ func DetermineUseLiveData(logger golog.Logger, liveData *bool, sensors []string)
...
+	}
+
+	// TODO Replace this config_path with a more sensible value
+	_, err = attrCfg.Validate("config_path")
+	if err != nil {
+		return &AttrConfig{}, newError(err.Error())
+	}
+
+	return attrCfg, nil
+}
```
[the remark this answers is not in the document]
tessavitabile: Yes, let's assume RSDK-1820 will be done.

config/config_test.go
```diff
@@ -70,3 +73,115 @@ func TestDetermineUseLiveData(t *testing.T) {
...
+func makeCfgService() config.Service {
+	model := resource.NewDefaultModel(resource.ModelName("test"))
+	cfgService := config.Service{Name: "test", Type: "slam", Model: model}
+	cfgService.Attributes = make(map[string]interface{})
+	cfgService.Attributes["config_params"] = map[string]string{
+		"mode": "test mode",
+	}
+	cfgService.Attributes["data_dir"] = "path"
+	cfgService.Attributes["use_live_data"] = true
+	cfgService.Attributes["delete_processed_data"] = true
```
tessavitabile, line 87: delete_processed_data isn't a required parameter, so it shouldn't be included in the simplest possible config.

config/config_test.go
```diff
@@ -70,3 +73,115 @@ func TestDetermineUseLiveData(t *testing.T) {
...
+		cfgService := makeCfgService()
+		cfgService.Attributes["use_live_data"] = "foo"
+		_, err := NewAttrConfig(cfgService)
+		test.That(t, err, test.ShouldBeError)
+		cfgService.Attributes["use_live_data"] = true
+		_, err = NewAttrConfig(cfgService)
+		test.That(t, err, test.ShouldBeNil)
+	})
+	/*
+	   TODO link to JIRA ticket that should have this fail
```
tessavitabile, line 126: This is https://viam.atlassian.net/browse/RSDK-1990. But we should probably have our own ticket (dependent on that one), since the SDK ticket probably won't cover the slam repo.

config/config.go
```diff
@@ -39,3 +41,96 @@ func DetermineUseLiveData(logger golog.Logger, liveData *bool, sensors []string)
...
+
+	if config.UseLiveData == nil {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "use_live_data")
+	}
+
+	if config.DeleteProcessedData == nil {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "delete_processed_data")
+	}
+
+	if config.DataRateMs != 0 && config.DataRateMs < 0 {
```
tessavitabile, line 95: In RuntimeConfigValidation(), we also currently verify that DataRateMs is at least minDataRateMs (though this is going away in https://viam.atlassian.net/browse/RSDK-1922)

tessavitabile (changes requested):
Great work! A few comments. I also agree with Kat's suggestion to wait to finish this until after RSDK-1820.

config/config.go
```diff
@@ -39,3 +41,96 @@ func DetermineUseLiveData(logger golog.Logger, liveData *bool, sensors []string)
 	}
 	return useLiveData, nil
 }
+
+// AttrConfig describes how to configure the service.
```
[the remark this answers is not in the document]
zaporter-work: Will there be lots of other logic that goes into config.go? Right now without attrconfig, config.go is pretty bare.

config/config.go
kkufieta: Can't we combine RSDK-1922 with this work? We can get rid of a fixed `minDataRateMs` now. Instead of using the hardcoded `minDataRateMs`, we can overwrite this value in orbslam only if needed. But it shouldn't be applied generically to all slam libraries.

tessavitabile: Yes, that makes sense. I had thought of removing the limit when we were ready to document suggested data rates for different setups. But I think it's doing more harm than good currently.

config/config.go
```diff
@@ -39,3 +41,96 @@ func DetermineUseLiveData(logger golog.Logger, liveData *bool, sensors []string)
...
+type AttrConfig struct {
+	Sensors             []string          `json:"sensors"`
+	ConfigParams        map[string]string `json:"config_params"`
+	DataDirectory       string            `json:"data_dir"`
+	UseLiveData         *bool             `json:"use_live_data"`
+	DataRateMs          int               `json:"data_rate_msec"`
+	MapRateSec          *int              `json:"map_rate_sec"`
```
kkufieta, line 52: [nit] Let's change `DataRateMs` to reflect how it's defined in JSON and how `MapRateSec` is implemented

```suggestion
	DataRateMsec          int               `json:"data_rate_msec"`
	MapRateSec          *int              `json:"map_rate_sec"`
```

config/config_test.go
tessavitabile: thank you!

zaporter-work:
SetParameters(...) feels like a non-intuitive name for what the function does. Would something like SetDefaultParameters(..), SetUnpassedParameters(..), UpdateConfigWithDefaults(..), .. work better?

config/config.go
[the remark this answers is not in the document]
kkufieta: No :) Seems like keeping it in here makes most sense then!

kkufieta:
> [nit] SetParameters(...) feels like a non-intuitive name for what the function does. I used this name because that is what the ticket suggested. However, would something like SetDefaultParameters(..), SetUnpassedParameters(..), UpdateConfigWithDefaults(..), .. work better?

@zaporter-work Agreed, let's choose a different name. Maybe `SetDefaultParameters`? I don't feel strongly, though, so you should pick what you prefer.

zaporter-work:
@tessavitabile Kat and I were talking about RSDK-1820 this morning and despite the great progress on that ticket, we are worried that this smaller detail is slowing down forward progress. And as long as you are okay with it, we think it might increase velocity if we just use this hacky solution for now and then create a ticket to fix it ASAP once that PR is done.

Another option is just to rebase everything on top of this PR and then continue work (what I am currently doing to work on RSDK-2004). However we might have to do this for a few PRs because while the other tickets aren't strictly blocked by this, many tests have implicit dependencies on AttrConf. 

@kkufieta If I have misunderstood / misrepresented anything you said, please feel free to correct me.

tessavitabile:
Yes, that sounds fine with me, as long as we don't end up adding "depends_on" to the config and then taking it out (but I think we won't because implicit dependencies will be done before remove the builtin slam service from rdk). Is this PR ready for re-review now?

config/config.go
```diff
@@ -39,3 +41,92 @@ func DetermineUseLiveData(logger golog.Logger, liveData *bool, sensors []string)
...
+
+// AttrConfig describes how to configure the SLAM service.
+type AttrConfig struct {
+	Sensors             []string          `json:"sensors"`
+	ConfigParams        map[string]string `json:"config_params"`
+	DataDirectory       string            `json:"data_dir"`
+	UseLiveData         *bool             `json:"use_live_data"`
+	DataRateMsec        int               `json:"data_rate_msec"`
+	MapRateSec          *int              `json:"map_rate_sec"`
+	InputFilePattern    string            `json:"input_file_pattern"`
```
tessavitabile, line 53: This has already been removed from rdk

config/config.go
```diff
@@ -39,3 +41,92 @@ func DetermineUseLiveData(logger golog.Logger, liveData *bool, sensors []string)
...
+	Port                string            `json:"port"`
+	DeleteProcessedData *bool             `json:"delete_processed_data"`
+	Dev                 bool              `json:"dev"`
```
tessavitabile, line 56: Please comment on https://viam.atlassian.net/browse/RSDK-1684, that the dev flag needs to be removed from the slam repo.

config/config.go
```diff
@@ -39,3 +41,92 @@ func DetermineUseLiveData(logger golog.Logger, liveData *bool, sensors []string)
...
+}
+
+// NewAttrConfig creates a SLAM config from a service config.
+func NewAttrConfig(cfg config.Service) (*AttrConfig, error) {
```
tessavitabile, line 60: Please create a ticket for how this function will get removed/modified after https://viam.atlassian.net/browse/RSDK-1820.

config/config.go
```diff
@@ -39,3 +41,92 @@ func DetermineUseLiveData(logger golog.Logger, liveData *bool, sensors []string)
...
+
+	if config.DataDirectory == "" {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "data_dir")
+	}
+
+	if config.UseLiveData == nil {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "use_live_data")
+	}
+
+	if config.DataRateMsec != 0 && config.DataRateMsec < 0 {
```
tessavitabile, line 91: Can remove the first check from the if statement

config/config_test.go
```diff
@@ -70,3 +73,131 @@ func TestDetermineUseLiveData(t *testing.T) {
...
+		test.That(t, err, test.ShouldBeError)
+	})
+
+	t.Run("Simplest valid config", func(t *testing.T) {
+		cfgService := makeCfgService()
+		_, err := NewAttrConfig(cfgService)
+		test.That(t, err, test.ShouldBeNil)
+	})
+
+	t.Run("Config without required fields", func(t *testing.T) {
```
tessavitabile, line 107: Also test for missing mode

config/config_test.go
```diff
@@ -70,3 +73,131 @@ func TestDetermineUseLiveData(t *testing.T) {
...
+		cfgService := makeCfgService()
+		cfgService.Attributes["use_live_data"] = "true"
+		_, err := NewAttrConfig(cfgService)
+		test.That(t, err, test.ShouldBeError)
+		cfgService.Attributes["use_live_data"] = true
+		_, err = NewAttrConfig(cfgService)
+		test.That(t, err, test.ShouldBeNil)
+	})
+
+	/* Unused config parameters should throw an error or fail. See: RSDK-2076
```
tessavitabile, line 128: I think it's better to leave this test out. We won't forget to add a test as part of RSDK-2076.

tessavitabile (approved):
LGTM mod comments

config/config.go
zaporter-work: https://viam.atlassian.net/browse/RSDK-2155

config/config.go
```diff
@@ -39,3 +41,91 @@ func DetermineUseLiveData(logger golog.Logger, liveData *bool, sensors []string)
...
+func NewAttrConfig(cfg config.Service) (*AttrConfig, error) {
+	attrCfg := &AttrConfig{}
+
+	_, err := config.TransformAttributeMapToStruct(attrCfg, cfg.Attributes)
+	if err != nil {
+		return &AttrConfig{}, newError(err.Error())
+	}
+
+	// TODO Replace this config_path with a more sensible value
+	_, err = attrCfg.Validate("config_path")
```
kkufieta, line 68: [nit] 
```suggestion
	_, err = attrCfg.Validate("services.slam.attributes.fake")
```

kkufieta (approved):
LGTM mod nits

config/config.go
```diff
@@ -39,3 +41,92 @@ func DetermineUseLiveData(logger golog.Logger, liveData *bool, sensors []string)
...
+	deps := config.Sensors
+
+	return deps, nil
+}
+
+// SetOptionalParameters updates any unset optional config parameters to the values passed to this function.
+func (config *AttrConfig) SetOptionalParameters(port string, defaultDataRateMsec, defaultMapRateSec int, logger golog.Logger) error {
+	if config.Port == "" {
+		config.Port = port
+	}
```
kkufieta, line 107: @zaporter-work  [nit/opt] I still get tripped up by the name vs. the fact that arguments are (only partially) prepended with `default...`. Can we rename `port` --> `defaultPort` to make it more consistent?

kkufieta: If there's an argument not to, that's fine. It's really opt or super nit.

zaporter-work: Done. Good catch.

config/config_test.go
```diff
@@ -70,3 +73,127 @@ func TestDetermineUseLiveData(t *testing.T) {
...
+		cfgService := makeCfgService()
+		cfgService.Attributes["use_live_data"] = "true"
+		_, err := NewAttrConfig(cfgService)
+		test.That(t, err, test.ShouldBeError)
+		cfgService.Attributes["use_live_data"] = true
+		_, err = NewAttrConfig(cfgService)
+		test.That(t, err, test.ShouldBeNil)
+	})
+
+	t.Run("Config with unacceptable values", func(t *testing.T) {
```
nicksanford, line 135:
```suggestion
	t.Run("Config with out of range values", func(t *testing.T) {
```

nicksanford (approved):

PR #160, opened 2023-02-23
RSDK-2009 - Add initial sensors implementation

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-2009

* Implements sensors.go
* Implements depth.go, lidar.go, rgb.go

**NOTE:**
This implementation is not expected to be functional and will likely change. It is a prerequisite for the following tickets because it sets the stage for refactoring `builtin.go` into model implementations for cartographer & orbslam:

* RSDK-2014 - [viam-cartographer] Implement cartographer model
* RSDK-2015 - [viam-orb-slam3] Implement orbslam3 model

The code here will likely change based on how the work on RSDK-2014 & RSDK-2015 utilizes it. You can see an example of how the sensors could be used in the following code snippets:
* https://gist.github.com/kkufieta/4e225301ea97077517fb58450612fa1e
* https://gist.github.com/kkufieta/24583b4e6e64c0bfb254a32acc86f1ff

Because the code in this PR is supposed to be preliminary work and is likely to change, I created a [separate ticket for writing tests](https://viam.atlassian.net/browse/RSDK-2201), which will happen either during or after RSDK-2014 & RSDK-2015.


sensors/lidar/lidar.go
```diff
@@ -0,0 +1,53 @@
...
+// New creates a new Lidar sensor based on the sensor definition and the service config.
+func New(ctx context.Context, deps registry.Dependencies, sensor sensors.Sensor, svcConfig *config.AttrConfig) (Lidar, error) {
+	name, err := sensor.GetName(svcConfig)
+	if err != nil {
+		return Lidar{}, err
+	}
+
+	newLidar, err := camera.FromDependencies(deps, name)
+	if err != nil {
+		return Lidar{}, errors.Wrapf(err, "error getting camera %v for slam service", name)
```
zaporter-work, line 32: Do we want to start distinguishing between a camera and a sensor (like lidar)?

sensors/sensors.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/pkg/errors"
+
+	"go.viam.com/slam/config"
+)
+
+// Sensor contains the necessary information to define a generic sensor.
+type Sensor struct {
+	ID                string
+	Index             int
+	DefaultDataRateMs int
```
zaporter-work, line 15:
```suggestion
	DefaultDataRateMsec int
```

sensors/depth/depth.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	return Depth{
+		Name:         name,
+		depth:        newDepth,
+		DataRateMsec: sensor.GetDataRateMsec(svcConfig),
+	}, nil
+}
+
+// GetData returns data from the depth sensor.
```
zaporter-work, line 46: Do we want to include more information about this very complex type signature? From the type signature alone, it is not clear what the func() return value is. Yet, it is important that the user releases the image.

sensors/rgb/rgb.go
```diff
@@ -0,0 +1,81 @@
...
+
+	return RGB{
+		Name:         name,
+		rgb:          newRGB,
+		DataRateMsec: sensor.GetDataRateMsec(svcConfig),
+	}, nil
+}
+
+// GetData returns data from the RGB sensor.
+func (rgb RGB) GetData(ctx context.Context) ([]byte, func(), error) {
```
zaporter-work, line 48: Ditto

sensors/utils/utils.go
```diff
@@ -0,0 +1,42 @@
...
+// function expects it to be an YCBCR image. If it is neither a lazy PNG nor an YCBCR, the
+// function errors out.
+func GetPNGImage(ctx context.Context, cam camera.Camera) ([]byte, func(), error) {
+	// We will hint that we want a PNG.
+	// The Camera service server implementation in RDK respects this; others may not.
+	readImgCtx := gostream.WithMIMETypeHint(ctx, utils.WithLazyMIMEType(utils.MimeTypePNG))
+	img, release, err := camera.ReadImage(readImgCtx, cam)
+	if err != nil {
+		return nil, nil, err
+	}
```
zaporter-work, line 25: Are we concerned about some program ignoring these errors and continuing to read images? Would that cause a memory leak?

zaporter-work (approved):
LGTM mod nits

sensors/depth/depth.go
```diff
@@ -0,0 +1,53 @@
...
+	"go.viam.com/slam/config"
+	"go.viam.com/slam/sensors"
+	"go.viam.com/slam/sensors/utils"
+)
+
+// Depth represents a depth sensor.
+type Depth struct {
+	Name         string
+	depth        camera.Camera
+	DataRateMsec int
```
tessavitabile, line 20: Why is this part of the sensor, rather than the data process?

sensors/depth/depth.go
```diff
@@ -0,0 +1,53 @@
...
+	"go.viam.com/rdk/components/camera"
+	"go.viam.com/rdk/registry"
+
+	"go.viam.com/slam/config"
+	"go.viam.com/slam/sensors"
+	"go.viam.com/slam/sensors/utils"
+)
+
+// Depth represents a depth sensor.
+type Depth struct {
```
tessavitabile, line 17: Do we need type validation that Depth inherits from Sensor?

sensors/sensors.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"github.com/pkg/errors"
+
+	"go.viam.com/slam/config"
+)
+
+// Sensor contains the necessary information to define a generic sensor.
+type Sensor struct {
+	ID                string
```
tessavitabile, line 13: Do the individual Sensor types (Depth, etc) need to set ID, Index, and DefaultDataRateMs?

tessavitabile (changes requested):
Looks good! A few comments/questions

sensors/sensors.go
```diff
@@ -0,0 +1,33 @@
...
+	Index             int
```
nicksanford, line 14: What is the sensor index?

sensors/sensors.go
nicksanford, line 13: What is the sensor ID?

sensors/sensors.go
nicksanford: Also, how will this work when sensors is a dictionary?

sensors/depth/depth.go
nicksanford, line 20: Same question. I don't have a good understanding of how `depth|rgb|lidar|.DataRateMsec` & `sensor.DefaultDataRateMs` are intended to interact with one another & why this makes sense: https://github.com/viamrobotics/slam/pull/160/files#diff-255612532e71d064ed5e0725da530352d9e537d7410e071ace9aa15b3f7f8c36R28

sensors/utils/utils.go
```diff
@@ -0,0 +1,42 @@
...
+	if lazyImg, ok := img.(*rimage.LazyEncodedImage); ok {
+		if lazyImg.MIMEType() != utils.MimeTypePNG {
+			return nil, nil, errors.Errorf("expected mime type %v, got %T", utils.MimeTypePNG, img)
```
nicksanford, line 28: Why is this not `return nil, nil, errors.Errorf("expected mime type %v, got %T", utils.MimeTypePNG, lazyImg.MIMEType())`?

nicksanford (changes requested):

sensors/utils/utils.go
nicksanford: Actually, that is a really good point. I wonder if we should call that release function before returning an error...

sensors/depth/depth.go
kkufieta: That's a good point. I designed this with an older version of the movement scope doc in mind. I will remove this from the sensor and leave the ownership with the data process.

kkufieta: Will remove those! I designed this with an older version of the movement scope doc in mind.

sensors/depth/depth.go
kkufieta: `func()` is only only `!= nil` if `error == nil`. I'll add a comment, though, to clarify this.

sensors/lidar/lidar.go
kkufieta: Good point. I'll edit the comment to make this clearer.

sensors/sensors.go
kkufieta: The index refers to the position of the sensor in the `sensors` array. If we change `sensors` to be a dictionary, this will be replaced with an `ID`.

sensors/sensors.go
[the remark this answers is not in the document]
kkufieta: This is a good point. I designed this with an old version of the movement scope in mind. I will remove the default data rate from the sensor type.

sensors/utils/utils.go
kkufieta: `release` is always `nil` when `err != nil`. I'll add a comment to make this clear.

sensors/depth/depth.go
zaporter-work: But what is `func()`? From this type signature and comment alone, I have no idea.

sensors/utils/utils.go
zaporter-work: Would that still be `%T`? I think MimeType() returns a string, so I think you would change the `%T` to `%v`. Not confident though.

sensors/depth/depth.go
tessavitabile: I think I don't understand the relationship between Sensor and Depth, since I would expect Depth to inherit from Sensor since Depth is a type of Sensor. I would expect us to have a slice of Sensors of different types, e.g. Depth and RGB, which allows us to call GetData on each different Sensor according to its own implementation. How are Sensor and Depth going to be used?

sensors/depth/depth.go
kkufieta: It is a release function you must call once you're done using the image. As of Sept 27, 2022, it didn't do anything yet. I am not sure if that has changed since then or where this function is defined. @edaniels might know more.

sensors/utils/utils.go
kkufieta: I don't think we need to replace `img` with `lazyImg.MIMEType()`, since `img` will not be of mime type `png` if we get to this line.

sensors/utils/utils.go
kkufieta: I am always returning the `release` function now.

sensors/depth/depth.go
tessavitabile: Ah, I see.

Is anything else going to go into sensors.Sensor? If not, I think we don't need the type, since it is just a wrapper for an int Index. We could instead create a utility function to get the name from the index.

kkufieta: I am not sure I understand. Maybe it's best to discuss this in person. The main point of `sensors.Sensor` is that we can define all sensors in a file such as this one: https://gist.github.com/kkufieta/4e225301ea97077517fb58450612fa1e#file-sensors-go

In the sensors.go file, we can clearly see the connection between the sensor (e.g. `lidar2D`, `imu`, etc) and the index. I am not sure how we could achieve the same type of human readability without using a `sensors.Sensor` type, but am very interested to hear about alternative approaches.

Originally, the plan was for `sensor.Sensor` to work with the idea of using a dict for sensors, and it contained an ID, a min & max data rate, and a default data rate (all of these have been removed due to the latest changes in scope).

kkufieta: Recap from offline discussion: Will remove `sensors.Sensor` type and map variable names to indices directly, e.g. `lidar2D_index := 0`

sensors/utils/utils.go
```diff
@@ -0,0 +1,53 @@
...
+	"go.viam.com/slam/config"
+)
+
+// GetName returns the name of the sensor based on its index in the sensor array.
+func GetName(svcConfig *config.AttrConfig, index int) (string, error) {
+	if index > len(svcConfig.Sensors) {
+		return "", errors.New("index out of bounds")
+	}
+	return svcConfig.Sensors[index], nil
+}
```
kkufieta, line 23: @tessavitabile @EshaMaharishi I deleted `sensors.Sensor` and moved the `GetName` helper function into utils. Does this look like a sufficient solution?

sensors/depth/depth.go
kkufieta: Done!

sensors/utils/utils.go
tessavitabile: yes, looks good!

tessavitabile (approved):

sensors/depth/depth.go
```diff
@@ -0,0 +1,42 @@
...
+	Name  string
+	depth camera.Camera
+}
+
+// New creates a new Depth sensor based on the sensor definition and the service config.
+func New(ctx context.Context, deps registry.Dependencies, sensors []string, sensorIndex int) (Depth, error) {
```
EshaMaharishi, line 21: [note to self] We expect specific sensors to be at specific indexes in the config (e.g. the config in app.viam.com's config tab). We might want to replace this array with a dictionary.

EshaMaharishi, line 21: [note to self] This New function (and the ones for lidar and rgb) will approximately replace [this call](https://github.com/viamrobotics/rdk/blob/b4a5473abe9a4721ddfc5300c87927e52c7511fb/services/slam/builtin/builtin.go#L491), though builtin.go itself will be replaced with slam service implementations in the carto and orbslam repos.

sensors/depth/depth.go
```diff
@@ -0,0 +1,42 @@
...
+
+	return Depth{
+		Name:  name,
+		depth: newDepth,
+	}, nil
+}
+
+// GetData returns data from the depth sensor. The returned function is a release function
+// that must be called once the caller of GetData is done using the image.
+func (depth Depth) GetData(ctx context.Context) ([]byte, func(), error) {
```
EshaMaharishi, line 40: [note to self] This (and the one for rgb) will approximately replace the getPNGImage calls [currently in builtin.go](https://github.com/viamrobotics/rdk/blob/b4a5473abe9a4721ddfc5300c87927e52c7511fb/services/slam/builtin/builtin.go)

sensors/lidar/lidar.go
```diff
@@ -0,0 +1,42 @@
...
+
+	return Lidar{
+		Name:  name,
+		lidar: newLidar,
+	}, nil
+}
+
+// GetData returns data from the lidar sensor.
+func (lidar Lidar) GetData(ctx context.Context) (pointcloud.PointCloud, error) {
+	return lidar.lidar.NextPointCloud(ctx)
```
EshaMaharishi, line 41: [note to self] This will approximately replace [this](https://github.com/viamrobotics/rdk/blob/b4a5473abe9a4721ddfc5300c87927e52c7511fb/services/slam/builtin/builtin.go#L923) call

sensors/rgb/rgb.go
```diff
@@ -0,0 +1,78 @@
...
+	}, nil
+}
+
+// GetData returns data from the RGB sensor. The returned function is a release function
+// that must be called once the caller of GetData is done using the image.
+func (rgb RGB) GetData(ctx context.Context) ([]byte, func(), error) {
+	return utils.GetPNGImage(ctx, rgb.rgb)
+}
+
+func validate(ctx context.Context, rgb camera.Camera) error {
```
EshaMaharishi, line 49: This code came from [here](https://github.com/viamrobotics/rdk/blob/b4a5473abe9a4721ddfc5300c87927e52c7511fb/services/slam/builtin/builtin.go#L206-L240)

sensors/utils/utils.go
```diff
@@ -0,0 +1,51 @@
...
+		return "", errors.New("index out of bounds")
+	}
+	return sensors[index], nil
+}
+
+// GetPNGImage first attempts to get a lazy PNG image. If the image is not a lazy PNG, the
+// function expects it to be an YCBCR image. If it is neither a lazy PNG nor an YCBCR, the
+// function errors out. The returned function is a release function that must be called once
+// the caller of GetPNGImage is done using the image.
+func GetPNGImage(ctx context.Context, cam camera.Camera) ([]byte, func(), error) {
```
EshaMaharishi, line 27: [note to self] This code came from [here](https://github.com/viamrobotics/rdk/blob/b4a5473abe9a4721ddfc5300c87927e52c7511fb/services/slam/builtin/builtin.go#L780)

EshaMaharishi (approved):
LGTM!

PR #166, opened 2023-02-28
Define ABSL_LEGACY_THREAD_ANNOTATIONS

nicksanford:
- Define `ABSL_LEGACY_THREAD_ANNOTATIONS` in CMake pipeline so cartographer continues to build with absl version `20230125.1` and greater: https://github.com/abseil/abseil-cpp/releases.

I've tested this by running:
```bash
make cartoclean && make buf && make buildcarto && make installcarto
```

It compiles just fine on both.

I then ran cartogropher in mapping / updating mode on both mac & linux with those compiled grpc servers & they appeared to work on both platforms. I updated the map for about 5 min & rendered the map in the RDK control UI.

### DIFF
--- a/slam-libraries/Makefile
+++ b/slam-libraries/Makefile
@@ -95,7 +95,7 @@ testcarto:
 	cd viam-cartographer && ./scripts/test_cartographer.sh
 
 installcarto:
-	sudo cp ./viam-cartographer/build/carto_grpc_server /usr/local/bin/carto_grpc_server
+	sudo cp viam-cartographer/build/carto_grpc_server /usr/local/bin/carto_grpc_server
 
 installorb:
 	sudo cp viam-orb-slam3/bin/orb_grpc_server /usr/local/bin/orb_grpc_server
--- a/slam-libraries/Makefile
+++ b/slam-libraries/Makefile
@@ -94,6 +94,12 @@ install-lua-carto:
 testcarto:
 	cd viam-cartographer && ./scripts/test_cartographer.sh
 
+installcarto:
+	sudo cp ./viam-cartographer/build/carto_grpc_server /usr/local/bin/carto_grpc_server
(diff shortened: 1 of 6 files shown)


slam-libraries/Makefile
```diff
@@ -94,6 +94,12 @@ install-lua-carto:
 testcarto:
 	cd viam-cartographer && ./scripts/test_cartographer.sh
 
+installcarto:
+	sudo cp ./viam-cartographer/build/carto_grpc_server /usr/local/bin/carto_grpc_server
+
+installorb:
+	sudo cp viam-orb-slam3/bin/orb_grpc_server /usr/local/bin/orb_grpc_server
+
```
kkufieta, line 102: I like this, was actually wishing we had this just earlier today ^_^

kkufieta (approved):
Great find, thanks for fixing this!

slam-libraries/Makefile
```diff
@@ -94,6 +94,12 @@ install-lua-carto:
 testcarto:
 	cd viam-cartographer && ./scripts/test_cartographer.sh
 
+installcarto:
+	sudo cp ./viam-cartographer/build/carto_grpc_server /usr/local/bin/carto_grpc_server
```
tessavitabile, line 98: Why is there a dot in this one?

tessavitabile (approved):
To clarify, with this change, we will no longer need to pin a version of abseil, is that correct?

nicksanford:
> 

Yes that is correct.

slam-libraries/Makefile
nicksanford: fixed

nicksanford:
> To clarify, with this change, we will no longer need to pin a version of abseil, is that correct?

Yes that is correct.

PR #168, opened 2023-03-03
RSDK-2007 Add WriteBytesToFile and WritePCDToFile

zaporter-work:
https://viam.atlassian.net/browse/RSDK-2007

This brings over the code from builtin.go related to writing images and PCDs to files. It creates two main functions `WriteBytesToFile` and `WritePCDToFile`.


dataprocess/dataprocess.go
```diff
@@ -0,0 +1,75 @@
...
+		return lazyImg.RawData(), nil
+	}
+
+	if ycbcrImg, ok := img.(*image.YCbCr); ok {
+		pngImage, err := rimage.EncodeImage(ctx, ycbcrImg, rdkutils.MimeTypePNG)
+		if err != nil {
+			return nil, err
+		}
+		return pngImage, nil
+	}
```
tessavitabile, line 32: Need to check for not ok here

dataprocess/dataprocess.go
```diff
@@ -0,0 +1,75 @@
...
+	"image/png"
+	"os"
+
+	"github.com/pkg/errors"
+	pc "go.viam.com/rdk/pointcloud"
+	"go.viam.com/rdk/rimage"
+	rdkutils "go.viam.com/rdk/utils"
+)
+
+func convertImageToPNG(ctx context.Context, img image.Image) ([]byte, error) {
```
tessavitabile, line 18: We could consider moving convertImageToPNG and WriteImageToPNGFile into the viam-orb-slam3 repo and WritePCDToFile to the viam-cartographer repo, since they will only be used by those modules. In the full modularization project, we'll get rid of the data process for cartographer entirely. But I also think this is fine, and we can change this in full modularization if we want to.

tessavitabile (changes requested):

dataprocess/dataprocess.go
zaporter-work: The way this function is meant to work is to try multiple different ways to convert to a png. We only fail if all of them fail. !ok is fine, we just fall through to the next possible way to convert the image.

dataprocess/dataprocess.go
kkufieta: I don't think it's worth the effort to move it now since it'll be deleted soon, anyways.

dataprocess/dataprocess.go
zaporter-work: I will add a comment that explains this because I realize now that it is not clear.

kkufieta: I am not sure if we do. I think this might not be checking `ok` on purpose because if both cases (`lazyImg` & `ycbcrImg`) are not `ok`, then it'll default to `png.Encode(buf, img)`.

tessavitabile: In the original code, we fail if we got anything other than a lazy png or YCbCr, since this would be unexpected: https://github.com/viamrobotics/rdk/blob/0128690343e5718f04292048a939a7fbfca2120e/services/slam/builtin/builtin.go#L906. This was done because decoding a png from camera server and then re-encoding resulting in changes to the image that caused orbslam to be unable to use it for feature matching (https://viam.atlassian.net/browse/RSDK-1393). We wanted to explicitly error if we got something other than a lazy png (or YCbCr, for mono webcam), since otherwise we wouldn't see an explicit error from rdk, but orbslam would log that it is unable to track.

zaporter-work: Thank you for the clarification! I had no idea about the history of this function. I removed this option and added a comment that explains some of this.

dataprocess/dataprocess.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// convertImageToPNG only supports rimage.LazyPNG and image.YCbCr.
+// This was done because decoding a png from the camera server and then re-encoding
+// it results in changes to the image that caused orbslam to be unable to use it
+// for feature matching (RSDK-1393).
+func convertImageToPNG(ctx context.Context, img image.Image) ([]byte, error) {
+	if lazyImg, ok := img.(*rimage.LazyEncodedImage); ok {
+		if lazyImg.MIMEType() != rdkutils.MimeTypePNG {
+			return nil, errors.Errorf("expected mime type %v, got %T", rdkutils.MimeTypePNG, img)
```
tessavitabile, line 24: We should have a test for this case

dataprocess/dataprocess.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		return lazyImg.RawData(), nil
+	}
+
+	if ycbcrImg, ok := img.(*image.YCbCr); ok {
```
tessavitabile, line 29: We should have a test for this case

dataprocess/dataprocess.go
```diff
@@ -0,0 +1,73 @@
...
+		pngImage, err := rimage.EncodeImage(ctx, ycbcrImg, rdkutils.MimeTypePNG)
+		if err != nil {
+			return nil, err
+		}
+		return pngImage, nil
+	}
+	return nil, errors.New("Failed to convert image to PNG. It was not LazyPNG or YCbCr")
```
tessavitabile, line 36: We should have a test for this case

tessavitabile (approved):
LGTM mod comments

dataprocess/dataprocess.go
zaporter-work: Unfortunately, after all of this, Kat and I decided that this is the wrong place to be doing this image conversion. All of this code was saved but it will be going in a different PR and in a different file.

dataprocess/dataprocess.go
zaporter-work: Unfortunately, after all of this, Kat and I decided that this is the wrong place to be doing this image conversion. All of this code was saved but it will be going in a different PR and in a different file.

dataprocess/dataprocess.go
zaporter-work: Unfortunately, after all of this, Kat and I decided that this is the wrong place to be doing this image conversion. All of this code was saved but it will be going in a different PR and in a different file.

dataprocess/dataprocess.go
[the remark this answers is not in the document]
zaporter-work: Agreed. Deleting for now. We will bring this back in a later PR.

dataprocess/dataprocess.go
```diff
@@ -0,0 +1,38 @@
+// Package dataprocess manages code related to the data-accepting process
```
kkufieta, line 1: [opt] 

```suggestion
// Package dataprocess manages code related to the data-saving process
```

dataprocess/dataprocess.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"os"
+
+	pc "go.viam.com/rdk/pointcloud"
+)
+
+// WritePCDToFile Encodes the pointcloud and then saves it to the passed filename.
```
kkufieta, line 13: [nit] Reflect the style in RDK

```suggestion
// WritePCDToFile encodes the pointcloud and then saves it to the passed filename.
```

dataprocess/dataprocess.go
```diff
@@ -0,0 +1,38 @@
...
+func WritePCDToFile(ctx context.Context, pointcloud pc.PointCloud, filename string) error {
+	buf := new(bytes.Buffer)
+	err := pc.ToPCD(pointcloud, buf, 1)
+	if err != nil {
+		return err
+	}
+	return WriteBytesToFile(buf.Bytes(), filename)
+}
+
+// WriteBytesToFile Writes the passed bytes to the passed filename.
```
kkufieta, line 23: [nit] Reflect the style in RDK

```suggestion
// WriteBytesToFile writes the passed bytes to the passed filename.
```

dataprocess/dataprocess.go
```diff
@@ -0,0 +1,38 @@
...
+	"bufio"
+	"bytes"
+	"context"
+	"os"
+
+	pc "go.viam.com/rdk/pointcloud"
+)
+
+// WritePCDToFile Encodes the pointcloud and then saves it to the passed filename.
+func WritePCDToFile(ctx context.Context, pointcloud pc.PointCloud, filename string) error {
```
kkufieta, line 14: [nit] We need to pass `context.Context` only when it is used within the function (or passed along into other functions that use it).

```suggestion
func WritePCDToFile(pointcloud pc.PointCloud, filename string) error {
```

dataprocess/dataprocess.go
```diff
@@ -0,0 +1,38 @@
...
+	buf := new(bytes.Buffer)
+	err := pc.ToPCD(pointcloud, buf, 1)
+	if err != nil {
+		return err
+	}
+	return WriteBytesToFile(buf.Bytes(), filename)
```
kkufieta, line 20: Nice refactoring!

kkufieta (approved):
Great work! LGTM mod nits

dataprocess/dataprocess_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"os"
+	"testing"
+
+	"github.com/edaniels/golog"
+	pc "go.viam.com/rdk/pointcloud"
+	"go.viam.com/test"
+)
+
+func TestWriteBytesToFile(t *testing.T) {
+	_ = golog.NewTestLogger(t)
```
nicksanford, line 13: I recommend moving this into a t.Run so that people don't need to move it into a t.Run when a second test case is added.

dataprocess/dataprocess_test.go
```diff
@@ -0,0 +1,40 @@
...
+	err = WriteBytesToFile(actualBytes, fileDest)
+	test.That(t, err, test.ShouldBeNil)
+	// Test that the file was actually written
+	readBytes, err := os.ReadFile(fileDest)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, readBytes, test.ShouldResemble, actualBytes)
+}
+
+func TestWritePCDToFile(t *testing.T) {
+	_ = golog.NewTestLogger(t)
```
nicksanford, line 29: I recommend moving this into a t.Run so that people don't need to move it into a t.Run when a second test case is added.

nicksanford (approved):
Approved pending t.Run refactor

PR #171, opened 2023-03-07
RSDK-2006 - Refactor, rename, and move a few functions & variable names

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-2006

Blocks related PR in RDK repo: https://github.com/viamrobotics/rdk/pull/2012

Work done:
* Refactor & extract parts of `createTimestampFilenames` from RDK into SLAM repo's dataprocess/dataprocess.go
* Move `createKeyValuePairs` from RDK into utils/utils.go & rename it to `DictToString`
* Extract functions from RDK builtin_test.go into testhelper/testhelper.go


config/config.go
```diff
@@ -59,14 +59,12 @@ type AttrConfig struct {
 func NewAttrConfig(cfg config.Service) (*AttrConfig, error) {
 	attrCfg := &AttrConfig{}
 
-	_, err := config.TransformAttributeMapToStruct(attrCfg, cfg.Attributes)
-	if err != nil {
+	if _, err := config.TransformAttributeMapToStruct(attrCfg, cfg.Attributes); err != nil {
```
zaporter-work, line 62: [opt] I have avoided using this error-handling syntax in places like this because it makes the (already complicated) line much more complicated. Though, it might just be because I am new to reading go. Feel free to mark as resolved after you read this.

zaporter-work (approved):
I expected testhelper.go to have tests but I guess it is okay because it is expected to be used in many tests. 

Should be good to go with a few fixes.

config/config.go
tessavitabile: This is idiomatic error handling for go.

dataprocess/dataprocess_test.go
```diff
@@ -3,11 +3,27 @@ package dataprocess
...
+	"time"
 
+	"github.com/edaniels/golog"
 	pc "go.viam.com/rdk/pointcloud"
 	"go.viam.com/test"
 )
 
+func TestCreateTimestampFilename(t *testing.T) {
+	t.Run("Create filename with a timestamp", func(t *testing.T) {
+		_ = golog.NewTestLogger(t)
```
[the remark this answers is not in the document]
tessavitabile: I'm also curious

testhelper/testhelper.go
```diff
@@ -0,0 +1,102 @@
...
+
+func checkDataDirForExpectedFiles(t *testing.T, dir string, prev int, deleteProcessedData, useLiveData bool) (int, error) {
+	files, err := os.ReadDir(dir)
+	test.That(t, err, test.ShouldBeNil)
+
+	if prev == 0 {
+		return len(files), nil
+	}
+	if deleteProcessedData && useLiveData {
+		test.That(t, prev, test.ShouldBeLessThanOrEqualTo, dataBufferSize+1)
```
tessavitabile, line 90: I think this might have been wrong in the original code. I think this should be 
`test.That(t, len(files), test.ShouldBeLessThanOrEqualTo, prev+dataBufferSize+1)`

testhelper/testhelper.go
```diff
@@ -0,0 +1,102 @@
...
+		return 0
+	}
+}
+
+func checkDataDirForExpectedFiles(t *testing.T, dir string, prev int, deleteProcessedData, useLiveData bool) (int, error) {
+	files, err := os.ReadDir(dir)
+	test.That(t, err, test.ShouldBeNil)
+
+	if prev == 0 {
+		return len(files), nil
```
tessavitabile, line 87: I'm not sure why we returned early in the original code instead of performing the checks below. I think they would make sense even when prev == 0

tessavitabile (changes requested):
Looks good, but I think there might have been a couple issues in the original tests.

testhelper/testhelper.go
zaporter-work: I think that the idea is you pass prev=0 at the start to get a count of how many files are in the target directory, then perform some operation... then run the function again with the previous return value. That would explain why the function is the way it is.

tessavitabile (approved):

go.mod
```diff
@@ -38,6 +38,7 @@ require (
 	github.com/alingse/asasalint v0.0.11 // indirect
 	github.com/ashanbrown/forbidigo v1.3.0 // indirect
 	github.com/ashanbrown/makezero v1.1.1 // indirect
+	github.com/aybabtme/uniplot v0.0.0-20151203143629-039c559e5e7e // indirect
```
EshaMaharishi, line 41: [note to self] this is automated, you run `go mod tidy` and this gets updated

EshaMaharishi (approved):
LGTM mod minor comment

PR #182, opened 2023-03-20
Add error messages to sensors

kkufieta:


zaporter-work:
Why dont we do these checks in the `New(...)` function? It seems like it is invalid behavior to instantiate any of these sensors if these values are nil.

kkufieta:
> Why dont we do these checks in the New(...) function? It should be invalid behavior to instantiate any of these sensors if these values are nil (IMO).

I agree this is not ideal. This does only help with debugging in case someone ignores the error message returned by `New` and attempts to call `GetData` on the returned object anyways. I added this because I was writing tests - but I am unsure if testing for the case where one ignores the error message returned by `New` is necessary.

kkufieta:
@zaporter-work Actually, I am returning this when config has `sensors: []` in the `getLidar()` function:

```golang
	if len(svcConfig.Sensors) == 0 {
		logger.Debug("no sensor provided in 'sensors' config parameter")
		return lidar.Lidar{}, nil
	}
```

That's because `sensors: []` is currently a valid option (will change with modularization v2). Therefore, I think it's good for now to safeguard someone from using `GetData` when `Lidar == nil`.

zaporter-work (approved):
That seems reasonable.

EshaMaharishi:
Spoke offline: it would be a programmer error if nil was passed in, so I think a panic is okay. (If it had been a user input error, then it would be better to check for valid input rather than crash).

EshaMaharishi (approved):
LGTM with suggestion to remove the checks for nil

PR #186, opened 2023-04-12
Add warning to Readme & code, indicating that the code might be transient

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-2353

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 ## (In)stability Notice
 > **Warning**
 > This is an experimental feature. Stability is not guaranteed. Breaking changes are likely to occur, and occur often.
+> **Warning**
+> The code in this repo might be transient. Stability is not guaranteed.
 
 ## Overview
 This repo contains utils for implementing a [slam service](https://github.com/viamrobotics/rdk/blob/main/services/slam/slam.go) in Golang.
--- a/config/config.go
+++ b/config/config.go
@@ -1,4 +1,5 @@
 // Package config implements functions to assist with attribute evaluation in the SLAM service
+// Warning: The code in this package might be transient. Stability is not guaranteed.
 package config
 
 import (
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -3,6 +3,8 @@
 ## (In)stability Notice
 > **Warning**
 > This is an experimental feature. Stability is not guaranteed. Breaking changes are likely to occur, and occur often.
+> **Warning**
+> The code in this repo might be transient. Stability is not guaranteed.
```
EshaMaharishi, line 8: Looks like there's a similar warning above, can we omit this one?

config/config.go
```diff
@@ -1,4 +1,5 @@
 // Package config implements functions to assist with attribute evaluation in the SLAM service
+// Warning: The code in this package might be transient. Stability is not guaranteed.
```
EshaMaharishi, line 2: I wonder if the warning in the README would be sufficient... it will be one less thing to remember to do in any new files, and I'm not sure adding the warning to every file will provide more value.

kkufieta: That's true - though I wonder if we should leave it in for two reasons:
1. When we added "This is an Experimental package" to the original slam repo, we agreed to add it both to the readme and to the package description. 
2. Adding it in the package description will make it visible in go.viam.com in the package overview, for example like here: https://pkg.go.dev/go.viam.com/rdk@v0.2.33/services/slam/builtin
    * I think this is helpful for users who might search for help on certain functions and will be directed towards the documentation. They can then derive from the Overview that this package is not stable.

Given those reasons, do you still think that the README might be sufficient?

nicksanford (approved):
I didn't realize that adding each of these to the package ensure that the generated go docs also have the warning. This seems like a reasonable middle ground to me.

EshaMaharishi (approved):

